PR #10, opened 2025-05-14
MCP-2 Expose a tool to navigate my projects

nquinquenel:
- Added an HTTP client (for now with no customization for SSL & co)
- Removed the tool to fetch projects starting with a prefix and replaced it with a direct call with our HTTP client to `api/projects/search_my_project`

### DIFF
--- a/src/main/java/org/sonar/mcp/http/HttpClient.java
+++ b/src/main/java/org/sonar/mcp/http/HttpClient.java
@@ -1,3 +1,22 @@
+/*
+ * Sonar MCP Server
+ * Copyright (C) 2025 SonarSource
+ * sonarlint@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
 package org.sonar.mcp.http;
 
 import java.io.Closeable;
--- a/src/main/java/org/sonar/mcp/serverapi/EndpointParams.java
+++ b/src/main/java/org/sonar/mcp/serverapi/EndpointParams.java
@@ -1,7 +1,7 @@
 /*
- * SonarLint Core - Server API
- * Copyright (C) 2016-2025 SonarSource SA
- * mailto:info AT sonarsource DOT com
+ * Sonar MCP Server
+ * Copyright (C) 2025 SonarSource
+ * sonarlint@sonarsource.com
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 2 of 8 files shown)


src/main/java/org/sonar/mcp/SonarMcpServer.java
```diff
@@ -24,31 +24,58 @@
...
-import org.sonar.mcp.tools.FindIssuesTool;
+import org.sonar.mcp.tools.issues.SearchIssuesTool;
+import org.sonar.mcp.tools.projects.SearchMyProjectsTool;
+import org.sonar.mcp.tools.issues.AnalyzeIssuesTool;
+import org.sonarsource.sonarlint.core.serverapi.EndpointParams;
 
 public class SonarMcpServer {
+
   private static final McpLogger LOG = McpLogger.getInstance();
+  private static final String USER_AGENT = "Sonar MCP Server";
```
damien-urruty-sonarsource: It might be useful to also include the version. Also, do you know if there is a way to know the client executing the MCP server?

src/main/java/org/sonar/mcp/SonarMcpServer.java
```diff
@@ -24,31 +24,58 @@
...
+
+    var serverApi = initializeServerApi(mcpConfiguration);
+    var backendService = new BackendService(mcpConfiguration.getStoragePath(), mcpConfiguration.getPluginPath());
+
+    var findIssuesTool = new AnalyzeIssuesTool(backendService);
+    var searchMyProjectsTool = new SearchMyProjectsTool(serverApi);
+    var searchIssuesTool = new SearchIssuesTool(serverApi);
 
     var syncServer = McpServer.sync(transportProvider)
       .serverInfo(new McpSchema.Implementation("sonar-mcp-server", "0.0.1"))
```
damien-urruty-sonarsource: Could the version be read from the gradle configuration? From the Manifest maybe?

src/main/java/org/sonar/mcp/http/HttpClientAdapter.java
```diff
@@ -0,0 +1,157 @@
...
+      .build();
+    return executeAsync(httpRequest);
+  }
+
+  private static Response waitFor(CompletableFuture<Response> f) {
+    return f.join();
+  }
+
+  @Override
+  public AsyncRequest getEventStream(String url, HttpConnectionListener connectionListener, Consumer<String> messageConsumer) {
```
damien-urruty-sonarsource: Will we support server-sent events? We might get rid of that

src/main/java/org/sonar/mcp/serverapi/issues/IssuesApi.java
```diff
@@ -0,0 +1,100 @@
...
+  private String buildPath(@Nullable String[] projects) {
+    var path = new StringBuilder(SEARCH_PATH);
+    boolean hasQueryParams = false;
+
+    if (organization != null) {
+      path.append("?organization=").append(organization);
+      hasQueryParams = true;
+    }
+    if (projects != null && projects.length > 0) {
+      path.append(hasQueryParams ? "&" : "?").append("projects=").append(String.join(",", projects));
```
damien-urruty-sonarsource: We should urlEncode all query params

src/main/java/org/sonar/mcp/serverapi/projects/ProjectsApi.java
```diff
@@ -0,0 +1,59 @@
...
+    try (var response = helper.get(buildPath())) {
+      var responseStr = response.bodyAsString();
+      return new Gson().fromJson(responseStr, SearchResponse.class);
+    }
+  }
+
+  private static String buildPath() {
+    return SEARCH_MY_PROJECTS_PATH;
+  }
+
```
damien-urruty-sonarsource: Maybe a bit overkill to have a method for that

src/main/java/org/sonar/mcp/SonarMcpServer.java
nquinquenel: > do you know if there is a way to know the client executing the MCP server?

Negative

src/main/java/org/sonar/mcp/http/HttpClientAdapter.java
nquinquenel: For now I decided not to support it - which is why it's just throwing an exception. Can't really get rid of that, it's part of the interface

src/main/java/org/sonar/mcp/serverapi/issues/IssuesApi.java
nquinquenel: Very good point!

src/main/java/org/sonar/mcp/http/HttpClientAdapter.java
damien-urruty-sonarsource: Don't we also own the interface? We don't have to respect it I think, we can avoid implementing it in this project. Just to be clear, I'm fine with keeping it for now, it's just that it will annoy us for coverage

nquinquenel: I was leveraging SLCORE classes - I refactored mostly everything to remove any dependency there, and reimplemented everything on our side. At the same time, I removed unused things. Should be good

PR #32, opened 2025-05-25
MCP-28 Implement schema validation via unit testing

nquinquenel:
[MCP-28](https://sonarsource.atlassian.net/browse/MCP-28)

[MCP-28]: https://sonarsource.atlassian.net/browse/MCP-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/org/sonar/mcp/tools/McpSchemaValidationTest.java
+++ b/src/test/java/org/sonar/mcp/tools/McpSchemaValidationTest.java
@@ -307,10 +307,6 @@ class McpSchemaValidationTest {
   }
 
   private void validateJsonSchemaProperties(Map<String, Object> properties, String toolName) {
-    assertThat(properties)
-      .as("Properties should not be empty for tool '%s'", toolName)
-      .isNotEmpty();
-    
     for (var property : properties.entrySet()) {
       assertThat(property.getKey())
         .as("Property name should not be empty for tool '%s'", toolName)


src/test/java/org/sonar/mcp/tools/McpSchemaValidationTest.java
```diff
@@ -0,0 +1,347 @@
...
+  @ParameterizedTest(name = "Tool ''{0}'' should have a valid MCP schema")
+  @MethodSource("provideTools")
+  void tool_schema_should_be_valid_according_to_mcp_spec(Tool tool) {
+    var schema = tool.definition();
+    validateMcpToolSchema(schema);
+  }
+
+  @ParameterizedTest(name = "Tool ''{0}'' should follow MCP naming convention")
+  @MethodSource("provideTools")
+  void tool_name_should_follow_mcp_naming_convention(Tool tool) {
```
damien-urruty-sonarsource, line 69: Could you add an extra check in this test or a separate test to verify that the tool name complies with the telemetry allowed key names?

src/test/java/org/sonar/mcp/tools/McpSchemaValidationTest.java
```diff
@@ -0,0 +1,347 @@
...
+    assertThat(tool.definition().name())
+      .as("Tool name should not be null or empty")
+      .isNotNull()
+      .isNotEmpty()
+      .as("Tool name should follow snake_case convention (lowercase letters, numbers, underscores)")
+      .matches("^[a-z][a-z0-9_]*[a-z0-9]$")
+      .as("Tool name should not start or end with underscore")
+      .doesNotStartWith("_")
```
damien-urruty-sonarsource, line 77: Just a thought, we could make such checks when registering tools at runtime instead of in a unit test. If one of the check fails, we don't start the MCP server

src/test/java/org/sonar/mcp/tools/McpSchemaValidationTest.java
```diff
@@ -0,0 +1,347 @@
...
+              assertThat(enumValues)
+                .as("Enum property '%s' in tool '%s' should have enum values", property.getKey(), schema.name())
+                .isInstanceOf(Object[].class);
+              
+              var enumArray = (Object[]) enumValues;
+              assertThat(enumArray)
+                .as("Enum property '%s' in tool '%s' should have at least one value", property.getKey(), schema.name())
+                .isNotEmpty();
+              
+              // Validate enum values are strings (common case)
```
damien-urruty-sonarsource: This is not what you check below

src/test/java/org/sonar/mcp/tools/McpSchemaValidationTest.java
```diff
@@ -0,0 +1,347 @@
...
+          var type = propertyDef.get("type");
+          
+          assertThat(type)
+            .as("Property '%s' in tool '%s' should have a type", property.getKey(), schema.name())
+            .isNotNull();
+          
+          if (type instanceof String typeStr) {
+            assertThat(typeStr)
+              .as("Property '%s' type in tool '%s' should be a valid JSON Schema type", property.getKey(), schema.name())
+              .isIn("string", "number", "integer", "boolean", "array", "object", "null");
```
damien-urruty-sonarsource, line 245: I think this is mostly enforced by our current design (the builder), so it's not a very useful test. There is one missing thing in the builder though, for `addArrayProperty`, we could accept an enum for `itemsType` to ensure no wrong input can be passed. It's a bit similar to my remark about making such checks at runtime

nquinquenel: My idea was to use the types defined in the spec, and verify them during this test. We enforce them via our builder, but we could make a mistake there as well. Let's revisit this if needed 👀

PR #34, opened 2025-05-26
MCP-29 Avoid errors while sending telemetry

damien-urruty-sonarsource:
[MCP-29](https://sonarsource.atlassian.net/browse/MCP-29)

[MCP-29]: https://sonarsource.atlassian.net/browse/MCP-29?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonar/mcp/SonarMcpServer.java
+++ b/src/main/java/org/sonar/mcp/SonarMcpServer.java
@@ -79,6 +79,7 @@ public class SonarMcpServer {
       .tools(supportedTools.stream().map(this::toSpec).toArray(McpServerFeatures.SyncToolSpecification[]::new))
       .build();
     LOG.setOutput(syncServer);
+    backendService.initialize();
     Runtime.getRuntime().addShutdownHook(new Thread(() -> shutdown(syncServer, backendService)));
   }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/sonar/mcp/slcore/McpSonarLintRpcClient.java
```diff
@@ -173,7 +170,7 @@ public void didChangeMatchedSonarProjectBranch(String configScopeId, String newM
 
   @Override
   public TelemetryClientLiveAttributesResponse getTelemetryLiveAttributes() {
-    return null;
+    return new TelemetryClientLiveAttributesResponse(Map.of());
```
nquinquenel, line 173: There are some other client methods that expect a response object (with null parameters) that we currently don't provide. I'm not sure it has an impact but it could be safer to align as well.

damien-urruty-sonarsource: I checked them and there should be no impact for now. We will have to revisit anyway and implement a few for the analysis support

PR #41, opened 2025-05-31
MCP-41 Introduce new tool for fetching source and SCM info

nquinquenel:
[MCP-41](https://sonarsource.atlassian.net/browse/MCP-41)

[MCP-41]: https://sonarsource.atlassian.net/browse/MCP-41?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
@@ -19,7 +19,6 @@ package org.sonarsource.sonarqube.mcp;
 import io.modelcontextprotocol.server.McpServer;
 import io.modelcontextprotocol.server.McpServerFeatures;
 import io.modelcontextprotocol.server.McpSyncServer;
-import io.modelcontextprotocol.server.transport.StdioServerTransportProvider;
 import io.modelcontextprotocol.spec.McpSchema;
 import java.util.List;
 import java.util.Map;
@@ -45,6 +44,7 @@ import org.sonarsource.sonarqube.mcp.tools.rules.ShowRuleTool;
 import org.sonarsource.sonarqube.mcp.tools.rules.ListRuleRepositoriesTool;
 import org.sonarsource.sonarqube.mcp.tools.sources.GetRawSourceTool;
 import org.sonarsource.sonarqube.mcp.tools.sources.GetScmInfoTool;
+import org.sonarsource.sonarqube.mcp.transport.StdioServerTransportProvider;
 
 public class SonarQubeMcpServer {
 
--- a/src/main/java/org/sonarsource/sonarqube/mcp/transport/StdioServerTransportProvider.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/transport/StdioServerTransportProvider.java
@@ -1,0 +1,317 @@
+/*
+ * SonarQube MCP Server
+ * Copyright (C) 2025 SonarSource
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the Sonar Source-Available License Version 1, as published by SonarSource SA.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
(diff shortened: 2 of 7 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/tools/SchemaToolBuilder.java
```diff
@@ -60,6 +60,12 @@ public SchemaToolBuilder addRequiredStringProperty(String propertyName, String d
     return this;
   }
 
+  public SchemaToolBuilder addNumberProperty(String propertyName, String description) {
+    var content = Map.of(TYPE_PROPERTY_NAME, "number", DESCRIPTION_KEY_NAME, description);
```
damien-urruty-sonarsource, line 70: Not for this PR but for a later improvement, we could use an enum for the type, to make sure we don"t do any mistakes

src/main/java/org/sonarsource/sonarqube/mcp/tools/sources/GetScmInfoTool.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  private static Boolean parseBoolean(String value) {
+    return switch (value.toLowerCase(Locale.getDefault())) {
+      case "true", "yes" -> true;
+      case "false", "no" -> false;
+      default -> null;
+    };
+  }
```
damien-urruty-sonarsource: Also a later improvement maybe, couldn't this be part of the Arguments class? Something like `getOptionalBoolean` or `getOptionalStringAsBoolean`. By the way, I'm surprised there is no boolean type in the JSON spec, have you tried it?

src/test/java/org/sonarsource/sonarqube/mcp/tools/sources/GetRawSourceToolTests.java
```diff
@@ -0,0 +1,245 @@
...
+        package org.sonar.check;
+
+        public enum Priority {
+        /**
+          * WARNING : DO NOT CHANGE THE ENUMERATION ORDER
+          * the enum ordinal is used for db persistence
+          */
+          BLOCKER, CRITICAL, MAJOR, MINOR, INFO
+        }
+        """;
```
damien-urruty-sonarsource, line 223: Is there any special case to test for encoding? It is not returned by the Web API so I assume a fixed encoding is used, but it is well translated at the MCP transport layer?

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/sources/SourcesApi.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.gson.Gson;
+import javax.annotation.Nullable;
+import org.sonarsource.sonarqube.mcp.serverapi.ServerApiHelper;
+import org.sonarsource.sonarqube.mcp.serverapi.UrlBuilder;
+import org.sonarsource.sonarqube.mcp.serverapi.sources.response.ScmResponse;
+
+public class SourcesApi {
+
+  public static final String SOURCES_RAW_PATH = "/api/sources/raw";
+  public static final String SOURCES_SCM_PATH = "/api/sources/scm";
```
damien-urruty-sonarsource, line 28: I see an issue was raised here. I never saw a valid occurrence of this rule. I think it's worth raising the point to the sonar-java maintainers? Let me know if you want me to do it

src/main/java/org/sonarsource/sonarqube/mcp/tools/SchemaToolBuilder.java
nquinquenel: There's a unit test validating all the types. But I agree that we could put in place an Enum for better safety!

src/main/java/org/sonarsource/sonarqube/mcp/tools/sources/GetScmInfoTool.java
nquinquenel: There's a boolean in the spec, but you're right, it would probably be better.
I did this because of the API from SQ:C:

```
**Possible values**
true
false
yes
no
**Default value**
false
```

But we actually just care whether it's true or false. I will update this.

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/sources/SourcesApi.java
nquinquenel: I haven't found a valid use case for this rule in our code base either. Could be worth reporting this if you will.

PR #52, opened 2025-06-18
MCP-51 Fail at startup if SonarQube Server version is not supported

damien-urruty-sonarsource:


src/main/java/org/sonarsource/sonarqube/mcp/serverapi/system/response/StatusResponse.java
```diff
@@ -16,11 +16,8 @@
  */
 package org.sonarsource.sonarqube.mcp.serverapi.system.response;
 
-import javax.annotation.Nullable;
-
 public record StatusResponse(
```
nquinquenel, line 19: In the tool using this class, I think we have a bunch of null checks that should also be removed

src/test/java/org/sonarsource/sonarqube/mcp/SonarQubeVersionCheckerTest.java
```diff
@@ -0,0 +1,67 @@
...
+    versionChecker = new SonarQubeVersionChecker(serverApi);
+  }
+  @Test
+  void it_should_not_throw_if_sonarqube_cloud() {
+    when(serverApi.isSonarQubeCloud()).thenReturn(true);
+
+    assertThatCode(versionChecker::failIfSonarQubeServerVersionIsNotSupported)
+      .doesNotThrowAnyException();
+  }
+  @Test
```
nquinquenel:
```suggestion
  }
  
  @Test
```

src/test/java/org/sonarsource/sonarqube/mcp/SonarQubeVersionCheckerTest.java
```diff
@@ -0,0 +1,67 @@
...
+  private SonarQubeVersionChecker versionChecker;
+
+  @BeforeEach
+  void prepare() {
+    serverApi = mock(ServerApi.class);
+    systemApi = mock(SystemApi.class);
+    when(serverApi.systemApi()).thenReturn(systemApi);
+    versionChecker = new SonarQubeVersionChecker(serverApi);
+  }
+  @Test
```
nquinquenel:
```suggestion
  }
  
  @Test
```

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/system/Version.java
```diff
@@ -0,0 +1,63 @@
...
+  private Version(String version) {
+    this.name = version.trim();
+    var qualifierPosition = name.indexOf("-");
+    String nameWithoutQualifier;
+    if (qualifierPosition != -1) {
+      nameWithoutQualifier = name.substring(0, qualifierPosition);
+    } else {
+      nameWithoutQualifier = this.name;
+    }
+    final var split = nameWithoutQualifier.split("\\.");
```
nquinquenel: Is the `final` necessary? I think we have the habit not to write it

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/system/response/StatusResponse.java
damien-urruty-sonarsource: I missed that, thanks!

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/system/Version.java
damien-urruty-sonarsource: It's not. I copied part of this code from SLCORE and it was there. I removed it

damien-urruty-sonarsource:
@nquinquenel I forgot about SQCB, good catch. I adjusted the version check, it is a bit hacky but it should work for now until we get an answer from on-prem on the long-term solution

PR #63, opened 2025-07-01
MCP-65 Support empty env vars and system properties

damien-urruty-sonarsource:
[MCP-65](https://sonarsource.atlassian.net/browse/MCP-65)

[MCP-65]: https://sonarsource.atlassian.net/browse/MCP-65?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
@@ -116,17 +116,18 @@ public class McpServerLaunchConfiguration {
 
   @CheckForNull
   private static String getValueViaEnvOrPropertyOrDefault(Map<String, String> environment, String propertyName, @Nullable String defaultValue) {
-    var property = environment.get(propertyName);
-    if (property == null) {
-      property = System.getProperty(propertyName);
+    var value = environment.get(propertyName);
+    if (isNullOrBlank(value)) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
```diff
@@ -116,17 +116,18 @@ public boolean isSonarCloud() {
...
+      }
     }
-    if (property == null) {
-      property = defaultValue;
-    }
-    if (property != null && property.isBlank()) {
-      property = null;
-    }
-    return property;
+    return value;
```
nquinquenel, line 126: I was previously setting the value to null at the end if it's blank, because we check for nullity to know whether some variables are set or not later, shouldn't we keep this behavior?

damien-urruty-sonarsource: The last fallback in case of blank env var or system property is to pick the default value. They are already set to null when needed, so it should be OK? Maybe I miss something

nquinquenel: All good, I forgot the default value being null

PR #66, opened 2025-07-08
MCP-67 MCP-69 Fix show_rule and list_quality_gates

nquinquenel:
[MCP-67](https://sonarsource.atlassian.net/browse/MCP-67)
[MCP-69](https://sonarsource.atlassian.net/browse/MCP-69)

`show_rule`:

image

`list_quality_gates`:

image

[MCP-67]: https://sonarsource.atlassian.net/browse/MCP-67?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MCP-69]: https://sonarsource.atlassian.net/browse/MCP-69?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
@@ -17,22 +17,28 @@
 package org.sonarsource.sonarqube.mcp.serverapi.qualitygates.response;
 
 import java.util.List;
+import javax.annotation.Nullable;
 
 public record ListResponse(List<QualityGate> qualitygates, long default_id, Actions actions) {
 
   public record QualityGate(
-    // SQ:S only
-    Long id,
     String name,
-    Boolean isDefault,
-    Boolean isBuiltIn,
+    boolean isDefault,
+    boolean isBuiltIn,
     Actions actions,
-    // SQ:S only
+    // SQ:S only fields below
+    @Nullable
+    Long id,
+    @Nullable
     List<Condition> conditions,
     // SQ:C only fields below
+    @Nullable
     String caycStatus,
+    @Nullable
     Boolean hasStandardConditions,
+    @Nullable
     Boolean hasMQRConditions,
+    @Nullable
     Boolean isAiCodeSupported
   ) {
   }
--- a/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
@@ -20,7 +20,21 @@ import java.util.List;
 
 public record ListResponse(List<QualityGate> qualitygates, long default_id, Actions actions) {
 
-  public record QualityGate(long id, String name, boolean isDefault, boolean isBuiltIn, Actions actions, List<Condition> conditions) {
+  public record QualityGate(
+    // SQ:S only
+    Long id,
+    String name,
+    Boolean isDefault,
+    Boolean isBuiltIn,
+    Actions actions,
+    // SQ:S only
+    List<Condition> conditions,
+    // SQ:C only fields below
+    String caycStatus,
+    Boolean hasStandardConditions,
+    Boolean hasMQRConditions,
+    Boolean isAiCodeSupported
+  ) {
   }
 
   public record Actions(boolean rename, boolean setAsDefault, boolean copy, boolean associateProjects, boolean delete, boolean manageConditions) {
--- a/src/main/java/org/sonarsource/sonarqube/mcp/tools/qualitygates/ListQualityGatesTool.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/tools/qualitygates/ListQualityGatesTool.java
@@ -47,10 +47,10 @@ public class ListQualityGatesTool extends Tool {
 
     for (var gate : response.qualitygates()) {
       stringBuilder.append("\n").append(gate.name() != null ? gate.name() : "Unnamed");
-      if (Boolean.TRUE.equals(gate.isDefault())) {
+      if (gate.isDefault()) {
         stringBuilder.append(" [Default]");
       }
-      if (Boolean.TRUE.equals(gate.isBuiltIn())) {
+      if (gate.isBuiltIn()) {
         stringBuilder.append(" [Built-in]");
       }
       if (gate.id() != null) {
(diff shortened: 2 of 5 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
```diff
@@ -20,7 +20,21 @@
 
 public record ListResponse(List<QualityGate> qualitygates, long default_id, Actions actions) {
 
-  public record QualityGate(long id, String name, boolean isDefault, boolean isBuiltIn, Actions actions, List<Condition> conditions) {
+  public record QualityGate(
+    // SQ:S only
```
vnaskos-sonar: I think this comment may not be correct, id name... should be common for both. There is another comment bellow `// SQ:S only` 🤔

src/main/java/org/sonarsource/sonarqube/mcp/tools/qualitygates/ListQualityGatesTool.java
```diff
@@ -46,25 +46,54 @@ private static String buildResponseFromList(ListResponse response) {
...
-        stringBuilder.append("Conditions:\n");
-        for (var condition : gate.conditions()) {
-          stringBuilder.append("- ").append(condition.metric())
-            .append(" ").append(condition.op())
-            .append(" ").append(condition.error())
-            .append("\n");
-        }
-      } else {
-        stringBuilder.append("No conditions\n");
+      stringBuilder.append("\n").append(gate.name() != null ? gate.name() : "Unnamed");
```
vnaskos-sonar: the String building part could be moved to the `toString` override method in the `QualityGate` record.

vnaskos-sonar: Instead of `gate.name() != null ? gate.name() : "Unnamed"`, you can use `Objects.requireNonNullElse(name, "Unnamed")`

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
```diff
@@ -20,7 +20,21 @@
...
+    Long id,
+    String name,
+    Boolean isDefault,
+    Boolean isBuiltIn,
+    Actions actions,
+    // SQ:S only
```
damien-urruty-sonarsource: Is this comment needed? There is a similar one a few lines above, so removing it would mean it's still for SQ:S only

src/main/java/org/sonarsource/sonarqube/mcp/tools/qualitygates/ListQualityGatesTool.java
```diff
@@ -46,25 +46,54 @@ private static String buildResponseFromList(ListResponse response) {
...
+    if (gate.conditions() == null) {
+      return;
+    }
+    if (!gate.conditions().isEmpty()) {
+      sb.append("Conditions:\n");
+      for (var condition : gate.conditions()) {
+        sb.append("- ").append(condition.metric())
+          .append(" ").append(condition.op())
+          .append(" ").append(condition.error())
+          .append("\n");
```
vnaskos-sonar, line 79: That can be moved to the Condition#toString

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
```diff
@@ -20,7 +20,21 @@
 
 public record ListResponse(List<QualityGate> qualitygates, long default_id, Actions actions) {
 
-  public record QualityGate(long id, String name, boolean isDefault, boolean isBuiltIn, Actions actions, List<Condition> conditions) {
+  public record QualityGate(
+    // SQ:S only
+    Long id,
```
damien-urruty-sonarsource: Are these fields nullable then? We might want to annotate them to make this explicit

damien-urruty-sonarsource (approved):
LGTM, some minor comments

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
nquinquenel: The fields in-between are common for both SQ:S and SQ:C, it's really the field under the comment that is referenced. I can reorder them to avoid confusion

src/main/java/org/sonarsource/sonarqube/mcp/tools/qualitygates/ListQualityGatesTool.java
```diff
@@ -46,25 +46,54 @@ private static String buildResponseFromList(ListResponse response) {
...
+      if (Boolean.TRUE.equals(gate.isBuiltIn())) {
+        stringBuilder.append(" [Built-in]");
       }
+      if (gate.id() != null) {
+        stringBuilder.append(" (ID: ").append(gate.id()).append(")");
+      }
+      stringBuilder.append("\n");
+
+      appendConditions(stringBuilder, gate);
+      appendCloudFields(stringBuilder, gate);
```
vnaskos-sonar: To improve the readability StringBuilder can be replaced with java Text Blocks. In that case the the QualityGate name String building can also be extracted to a separate method.

```
String formattedConditions = String.join("\n", conditions.map(c => "-" c.toString());
String qualityGate = """
                     %s
                     Conditions:
                     %s"""
                     .formatted(formattedName(), formattedConditions)
```

vnaskos-sonar (approved):
All good, I just left some nitpicks 😉

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/qualitygates/response/ListResponse.java
nquinquenel: ID is only present in SQ:C response

src/main/java/org/sonarsource/sonarqube/mcp/tools/qualitygates/ListQualityGatesTool.java
nquinquenel: I like this proposition, however I'm not exactly sure we want to mix the actual response record with the logic of displaying text to the tool. The record might be used in different tools, and we typically don't display the full response object to the tool. We select specific information we want to expose because some of them have low values and might generate noise depending on the context (this could be challenged though).

PR #87, opened 2025-09-09
MCP-106 Improve support for parsing parameter types

nquinquenel:


src/main/java/org/sonarsource/sonarqube/mcp/tools/Tool.java
```diff
@@ -46,36 +46,53 @@ public String getStringOrThrow(String argumentName) {
       if (!argumentsMap.containsKey(argumentName)) {
         throw new MissingRequiredArgumentException(argumentName);
       }
-      return (String) argumentsMap.get(argumentName);
+      var arg = argumentsMap.get(argumentName);
+      if (arg instanceof String string) {
+        return string;
+      } else {
+        return String.valueOf(arg);
```
damien-urruty-sonarsource: This will return "null" if arg is null. I think we were returning null previously, which was probably wrong, so it should be fine

nquinquenel: Modified to throw when null

PR #94, opened 2025-09-18
MCP-124 MCP-117 MCP-116 Bridge and new tools for analysis with SQ:IDE

nquinquenel:
[MCP-124](https://sonarsource.atlassian.net/browse/MCP-124)
[MCP-117](https://sonarsource.atlassian.net/browse/MCP-117)
[MCP-116](https://sonarsource.atlassian.net/browse/MCP-116)

[MCP-124]: https://sonarsource.atlassian.net/browse/MCP-124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MCP-117]: https://sonarsource.atlassian.net/browse/MCP-117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MCP-116]: https://sonarsource.atlassian.net/browse/MCP-116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
@@ -34,9 +34,9 @@ import org.sonarsource.sonarqube.mcp.serverapi.ServerApiHelper;
 import org.sonarsource.sonarqube.mcp.slcore.BackendService;
 import org.sonarsource.sonarqube.mcp.tools.Tool;
 import org.sonarsource.sonarqube.mcp.tools.ToolExecutor;
-import org.sonarsource.sonarqube.mcp.tools.analysis.AutomaticAnalysisEnablementTool;
+import org.sonarsource.sonarqube.mcp.tools.analysis.ToggleAutomaticAnalysisTool;
 import org.sonarsource.sonarqube.mcp.tools.analysis.AnalysisTool;
-import org.sonarsource.sonarqube.mcp.tools.analysis.AnalyzeListFilesTool;
+import org.sonarsource.sonarqube.mcp.tools.analysis.AnalyzeFileListTool;
 import org.sonarsource.sonarqube.mcp.tools.enterprises.ListEnterprisesTool;
 import org.sonarsource.sonarqube.mcp.tools.issues.ChangeIssueStatusTool;
 import org.sonarsource.sonarqube.mcp.tools.issues.SearchIssuesTool;
@@ -94,8 +94,8 @@ public class SonarQubeMcpServer {
 
     if (sonarqubeIdeBridgeClient.isAvailable()) {
       LOG.info("SonarQube for IDE integration is available, enabling related tools.");
-      this.supportedTools.add(new AnalyzeListFilesTool(sonarqubeIdeBridgeClient));
-      this.supportedTools.add(new AutomaticAnalysisEnablementTool(sonarqubeIdeBridgeClient));
+      this.supportedTools.add(new AnalyzeFileListTool(sonarqubeIdeBridgeClient));
+      this.supportedTools.add(new ToggleAutomaticAnalysisTool(sonarqubeIdeBridgeClient));
     } else {
       this.supportedTools.add(new AnalysisTool(backendService, serverApi));
     }
--- a/src/main/java/org/sonarsource/sonarqube/mcp/bridge/SonarQubeIdeBridgeClient.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/bridge/SonarQubeIdeBridgeClient.java
@@ -31,8 +31,8 @@ public class SonarQubeIdeBridgeClient {
 
   private static final McpLogger LOG = McpLogger.getInstance();
 
-  public static final String AUTOMATIC_ANALYSIS_ENABLEMENT_PATH = "/sonarlint/api/analysis/automatic/config";
-  public static final String ANALYZE_LIST_FILES_PATH = "/sonarlint/api/analysis/files";
+  public static final String TOGGLE_AUTOMATIC_ANALYSIS_PATH = "/sonarlint/api/analysis/automatic/config";
+  public static final String ANALYSIS_FILES_PATH = "/sonarlint/api/analysis/files";
   public static final String STATUS_PATH = "/sonarlint/api/status";
 
   private final ServerApiHelper helper;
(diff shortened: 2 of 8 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/bridge/SonarQubeIdeBridgeClient.java
```diff
@@ -0,0 +1,103 @@
...
+      .serializeNulls()
+      .create();
+  }
+
+  public boolean isAvailable() {
+    try {
+      var response = helper.rawGet(STATUS_PATH);
+      return response.isSuccessful();
+    } catch (Exception e) {
+      LOG.info("SonarQube for IDE availability check failed");
```
damien-urruty-sonarsource: Maybe log the exception too? Else we will lose the cause forever

src/main/java/org/sonarsource/sonarqube/mcp/bridge/SonarQubeIdeBridgeClient.java
```diff
@@ -0,0 +1,103 @@
...
+  public record AnalyzeListFilesIssueResponse(
+    String ruleKey,
+    String message,
+    @Nullable String severity,
+    @Nullable String filePath,
+    @Nullable TextRange textRange
+  ) {
+  }
+
+  public record AutomaticAnalysisEnablementResponse(boolean success, String message) {
```
damien-urruty-sonarsource: I think success is no more present in the payload?

src/main/java/org/sonarsource/sonarqube/mcp/bridge/SonarQubeIdeBridgeClient.java
```diff
@@ -0,0 +1,103 @@
...
+    try (var response = helper.post(ANALYZE_LIST_FILES_PATH, HttpClient.JSON_CONTENT_TYPE, requestBody)) {
+      var analysisResponse = gson.fromJson(response.bodyAsString(), AnalyzeListFilesResponse.class);
+      return Optional.of(analysisResponse);
+    } catch (Exception e) {
+      LOG.error("Error requesting file analysis", e);
+      return Optional.empty();
+    }
+  }
+
+  public record AnalyzeListFilesRequest(List<String> fileList) {
```
damien-urruty-sonarsource: Isn't it `fileAbsolutePaths` now?

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeFileListTool.java
```diff
@@ -0,0 +1,107 @@
...
+        "This tool connects to a running SonarQube for IDE instance to perform code quality analysis on a list of files.")
+      .addArrayProperty(LIST_FILES_PROPERTY, "string", "List of absolute file paths to analyze")
+      .build());
+    this.bridgeClient = bridgeClient;
+  }
+
+  @Override
+  public Result execute(Arguments arguments) {
+    if (!bridgeClient.isAvailable()) {
+      return Result.failure("SonarQube for IDE is not available. Please ensure SonarQube for IDE is running.");
```
damien-urruty-sonarsource, line 45: As configuration is necessary, it should also be part of this message? Like 'is running and the bridge is configured with...'

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeListFilesTool.java
```diff
@@ -0,0 +1,107 @@
...
+  @Override
+  public Result execute(Arguments arguments) {
+    if (!bridgeClient.isAvailable()) {
+      return Result.failure("SonarQube for IDE is not available. Please ensure SonarQube for IDE is running.");
+    }
+
+    var listFiles = arguments.getOptionalStringList(LIST_FILES_PROPERTY);
+    if (listFiles == null || listFiles.isEmpty()) {
+      return Result.failure("No files provided to analyze. Please provide a list of file paths using the '" + LIST_FILES_PROPERTY + "' property.");
+    }
```
damien-urruty-sonarsource: It makes it look like the arg is optional, but it isn't in fact. This could be moved to the Arguments class, e.g. `getNonEmptyStringListOrThrow()`.

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
```diff
@@ -87,6 +90,13 @@ public SonarQubeMcpServer(StdioServerTransportProvider transportProvider, Map<St
     this.sonarQubeVersionChecker = new SonarQubeVersionChecker(serverApi);
     this.pluginsSynchronizer = new PluginsSynchronizer(serverApi, mcpConfiguration.getStoragePath());
     this.toolExecutor = new ToolExecutor(backendService);
+    var sonarqubeIdeBridgeClient = initializeBridgeClient(mcpConfiguration);
+
+    if (sonarqubeIdeBridgeClient.isAvailable()) {
+      LOG.info("SonarQube for IDE integration is available, enabling related tools.");
+      this.supportedTools.add(new AnalyzeListFilesTool(sonarqubeIdeBridgeClient));
```
damien-urruty-sonarsource: Do we still keep the other analysis tool if this one is active?

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AutomaticAnalysisEnablementTool.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public Result execute(Arguments arguments) {
+    if (!bridgeClient.isAvailable()) {
+      return Result.failure("SonarQube for IDE is not available. Please ensure SonarQube for IDE is running.");
+    }
+
+    var enabled = arguments.getBooleanOrThrow(ENABLED_PROPERTY);
+
+    var updateResult = bridgeClient.requestAutomaticAnalysisEnablement(enabled);
```
damien-urruty-sonarsource: Something that I just think about now. What if the user already had automatic analysis disabled as part of its preferences when working in the IDE? The agent might end up re-enabling it? Could we avoid that?

damien-urruty-sonarsource (commented):
If I'm not wrong the payloads are not up-to-date compared to the merged implementation in SLCORE. I guess we would also need integration tests for that at some point, but it's not a trivial thing to setup

src/main/java/org/sonarsource/sonarqube/mcp/bridge/SonarQubeIdeBridgeClient.java
nquinquenel: You're right, I just did a live check and fixed the discrepancies for both tools....

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeFileListTool.java
nquinquenel: The tool should be added only if SonarQube for IDE was detected as available at start up, so I suppose the user already had the configuration set up at some point

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeListFilesTool.java
nquinquenel: You're right, I'll use `getStringListOrThrow`, but I'd prefer to avoid having another method for args 🤷🏻

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
nquinquenel: Yes good point, I also thought the same and I just disabled the other tool, to avoid conflicts

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AutomaticAnalysisEnablementTool.java
nquinquenel: Mmh I think it's up to the user to decide when to call this. If he does not want to re-anable analysis then he won't call the tool. Ideally the behavior will be described in a rule file that the user can customize. I think that even if you had the automatic analysis disabled in the settings, you may want to re-enable it for whatever reason?

nquinquenel:
Validated that it works:
image
image

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AutomaticAnalysisEnablementTool.java
damien-urruty-sonarsource: I don't know. I am just worried to change user configuration without user interaction, because if we create a rule file automatically, the user won't necessarily know what happens behind the scenes

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeListFilesTool.java
```diff
@@ -0,0 +1,107 @@
...
+import org.sonarsource.sonarqube.mcp.log.McpLogger;
+import org.sonarsource.sonarqube.mcp.bridge.SonarQubeIdeBridgeClient;
+import org.sonarsource.sonarqube.mcp.tools.SchemaToolBuilder;
+import org.sonarsource.sonarqube.mcp.tools.Tool;
+
+public class AnalyzeListFilesTool extends Tool {
+  private static final McpLogger LOG = McpLogger.getInstance();
+
+  public static final String TOOL_NAME = "analyze_list_files";
+  public static final String LIST_FILES_PROPERTY = "list_files";
```
damien-urruty-sonarsource: I had missed the toll name in the previous review. `analyze_list_files` sounds strange, I would suggest `analyze_files_list` or `analyze_files`.
Similarly, `list_files` seems also odd, maybe `files_list` or `files`, or `file_absolute_paths` if we want to be clearer

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AutomaticAnalysisEnablementTool.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+package org.sonarsource.sonarqube.mcp.tools.analysis;
+
+import org.sonarsource.sonarqube.mcp.bridge.SonarQubeIdeBridgeClient;
+import org.sonarsource.sonarqube.mcp.tools.SchemaToolBuilder;
+import org.sonarsource.sonarqube.mcp.tools.Tool;
+
+public class AutomaticAnalysisEnablementTool extends Tool {
+
+  public static final String TOOL_NAME = "automatic_analysis_enablement";
```
damien-urruty-sonarsource: All tool names are general verbs, to show that it's an action. What about `toggle_analysis_enablement` or `toggle_automatic_analysis`?

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeListFilesTool.java
nquinquenel: For the past 2 weeks I was thinking about this name and that something was off... I changed to `analyze_file_list` and `file_absolute_paths`

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AutomaticAnalysisEnablementTool.java
nquinquenel: Right, I changed to `toggle_automatic_analysis`

PR #95, opened 2025-09-19
MCP-126 Publish MCP server into MCP registry

nquinquenel:
[MCP-126](https://sonarsource.atlassian.net/browse/MCP-126)

[MCP-126]: https://sonarsource.atlassian.net/browse/MCP-126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server.json
+++ b/server.json
@@ -1,0 +1,45 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
+  "name": "io.github.SonarSource/sonarqube-mcp-server",
+  "description": "An MCP server that enables integration with SonarQube Server or Cloud for code quality and security.",
+  "status": "active",
+  "repository": {
+    "url": "https://github.com/SonarSource/sonarqube-mcp-server",
+    "source": "github"
+  },
+  "version": "0.0.8",
+  "packages": [
+    {
+      "registry_type": "oci",
+      "registry_base_url": "https://docker.io",
+      "identifier": "mcp/sonarqube",
+      "version": "sha256:1f83c5655d462b267414e4b2e81863e3568ab2dc8c5e23ee094f4b1e061cbaf5",
+      "transport": {
+        "type": "stdio"
+      },
+      "environment_variables": [
+        {
+          "description": "Your SonarQube Server USER token",
+          "is_required": true,
+          "format": "string",
+          "is_secret": true,
+          "name": "SONARQUBE_TOKEN"
+        },
+        {
+          "description": "Your SonarQube Cloud organization key (if using SonarQube Cloud)",
+          "is_required": false,
+          "format": "string",
+          "is_secret": true,
+          "name": "SONARQUBE_ORG"
+        },
+        {
+          "description": "Your SonarQube Server URL (if using SonarQube Server)",
+          "is_required": false,
+          "format": "string",
+          "is_secret": true,
+          "name": "SONARQUBE_URL"
+        }
+      ]
+    }
+  ]
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,5 +46,6 @@ COPY --chown=appuser:appgroup --chmod=755 scripts/install-certificates.sh /usr/l
 USER appuser
 WORKDIR /app
 ENV STORAGE_PATH=./storage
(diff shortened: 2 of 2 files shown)


server.json
```diff
@@ -0,0 +1,45 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
+  "name": "io.github.SonarSource/sonarqube-mcp-server",
+  "description": "An MCP server that enables integration with SonarQube Server or Cloud for code quality and security.",
+  "status": "active",
+  "repository": {
+    "url": "https://github.com/SonarSource/sonarqube-mcp-server",
+    "source": "github"
+  },
+  "version": "0.0.8",
```
damien-urruty-sonarsource, line 10: Just a note, we will need to change the release procedure to bump this version. Apparently this file supports environment variables, see [here](https://github.com/domdomegg/airtable-mcp-server/blob/master/server.json#L10). Maybe we could keep the version in a single file to avoid having to bump several places, and have a script set the env var and call the publication command?

damien-urruty-sonarsource: We could also think about automating the release with a GitHub action

server.json
```diff
@@ -0,0 +1,45 @@
...
+  "packages": [
+    {
+      "registry_type": "oci",
+      "registry_base_url": "https://docker.io",
+      "identifier": "mcp/sonarqube",
+      "version": "sha256:1f83c5655d462b267414e4b2e81863e3568ab2dc8c5e23ee094f4b1e061cbaf5",
+      "transport": {
+        "type": "stdio"
+      },
```
damien-urruty-sonarsource, line 19: The declaration in this [example](https://github.com/domdomegg/airtable-mcp-server/blob/master/server.json#L30) is slightly different: 

> "registry_type": "oci",
> "identifier": "domdomegg/airtable-mcp-server",
> "version": "{{VERSION}}",
>  "runtime_hint": "docker",

Maybe it's fine like this, it's not clear which ones should be provided

server.json
```diff
@@ -0,0 +1,45 @@
...
+    "url": "https://github.com/SonarSource/sonarqube-mcp-server",
+    "source": "github"
+  },
+  "version": "0.0.8",
+  "packages": [
+    {
+      "registry_type": "oci",
+      "registry_base_url": "https://docker.io",
+      "identifier": "mcp/sonarqube",
+      "version": "sha256:1f83c5655d462b267414e4b2e81863e3568ab2dc8c5e23ee094f4b1e061cbaf5",
```
damien-urruty-sonarsource, line 16: The fact that we need a precise digest will be a bit annoying. When we want to release the server, we first need to merge the version update to master, wait max 6h for the new image with a new digest to be published on Docker MCP registry, then we will need to modify this file and release on GH MCP registry. For example, this digest will not be the correct one, as you added a LABEL in the Dockerfile, a new digest will be created. Nothing impossible, just a bit cumbersome

server.json
nquinquenel: What I pushed has been generated automatically via their `mcp-publisher` tool, and it also follows the given example on their documentation. I will keep an eye if something breaks due to this.

server.json
nquinquenel: Yes, this is definitely frustrating, and I’ve been thinking about it too. I don’t have a better solution in mind yet, and since we want to keep moving quickly, we’ll probably just need to put up with the longer release process for now. This digest will indeed be wrong, I just wanted to push the initial configuration.

The fact that we’re working inside this Docker environment, where we don’t have much control, makes everything more annoying 😮‍💨

server.json
nquinquenel: Ah good point, I missed this capability. I'll create a ticket about this for future hardening

server.json
nquinquenel: Given their schema spec: https://github.com/modelcontextprotocol/registry/blob/edd93af8ac7d8ecdd1d0799da64fe5776e2ac1e5/docs/reference/server-json/server.schema.json#L126

`runtime_hint` should be provided when `runtime_arguments` is. So I'm not sure if it's required in the example you shared 🤔

PR #103, opened 2025-10-01
MCP-140 Support the search_dependency_risks tool with SQC

damien-urruty-sonarsource:
[MCP-140](https://sonarsource.atlassian.net/browse/MCP-140)

[MCP-140]: https://sonarsource.atlassian.net/browse/MCP-140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
```diff
@@ -56,6 +58,7 @@ public McpServerLaunchConfiguration(Map<String, String> environment) {
     }
     this.storagePath = Paths.get(storagePathString);
     var sonarqubeCloudUrl = getValueViaEnvOrPropertyOrDefault(environment, SONARQUBE_CLOUD_URL, "https://sonarcloud.io");
+    this.sonarqubeCloudApiUrl = getValueViaEnvOrPropertyOrDefault(environment, SONARQUBE_CLOUD_API_URL, "https://api.sonarcloud.io");
```
vnaskos-sonar: This adds one more step for the users on configuring the MCP server. It's a bit complicated because it has:
1. to be manually provided by the user when the `SONARQUBE_CLOUD_URL` US instance is chosen. It has to be documented on the docs.
1. to match with the sonarqubeCloudUrl, eg. EU instance must have the EU API URL but we don't have a check for that.

I understand that we need the api url for the sca issues, but is there maybe an established convention to derive the api url from the sqc url? (something like prepend `api.` on sqc url).

Do we need both cloud url and cloud api url?

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
```diff
@@ -86,10 +86,10 @@ private HttpClient.Response rawPost(String url, String contentType, String body)
   }
 
   /**
-   * Execute GET using the API subdomain (api.sonarcloud.io)
+   * Execute GET using the API subdomain (api.sonarcloud.io) for SQC, or api/V2 for SQS
    */
-  public HttpClient.Response getApiSubdomain(String path) {
-    var response = rawGetApiSubdomain(path);
+  public HttpClient.Response getApiV2(String path) {
```
vnaskos-sonar: The name is misleading, SQC doesn't have a v2.

I am not sure if the name is the issue or the design in general, but I had really hard time understanding what is going on. 😕 

Part of the confusion is that we have multiple ways of checking if the instance is SQC or SQS, like `isSonarCloud` or `organization == null`. The whole application is aware of both and the `getApiV2` is trying to hide it.

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
```diff
@@ -99,30 +99,19 @@ public HttpClient.Response getApiSubdomain(String path) {
...
-    if (endpointParams.organization() == null) {
-      // For SonarQube Server, fall back to regular endpoint
-      return buildEndpointUrl(relativePath);
-    }
-    
-    var baseUrl = endpointParams.baseUrl();
-    // Transform sonarcloud.io to api.sonarcloud.io
-    if (baseUrl.contains("sonarcloud.io")) {
-      baseUrl = baseUrl.replace("://sonarcloud.io", "://api.sonarcloud.io");
-    }
```
vnaskos-sonar, line 123: Related to my previous comment about the urls. It seems here that we were calculating it. Although we take into account only the SQC EU production instance. We could keep that and try to make it more generic?

```
if (organization != null) {
   // replace https:// with https://api.
}
```

vnaskos-sonar (changes requested):
Feature-wise it works. However, the code is quite complicated with the new env variable and the v2 mechanism.

The PR has 3 topics
1. Add support for SQC SCA
1. Add support for more SQC instances (not only SQC EU)
1. Make the code aware of SQS v2 API

src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
damien-urruty-sonarsource: The last argument is the default value if the parameter is not set. So in 99% of cases, this new property does not need to be set, and the default value makes it work out of the box. You can see it more as an internal property, and I don't have the intention to document it. 

Deriving the API URL was done previously, but unfortunately it was not correct because it's not as simple as prepending a `api.`. Some SQC environments use a different convention, for example DevX instances are reachable via https://devx.sc-devx.io/ while the API is located at https://api.sc-devx.io/. As we expose a variable to let users change the SQC URL, it's more correct to also expose a variable to change the API URL. And it would be hard to check that both values are correct, so it's up to the user to configure them appropriately. But this is more advanced usage, probably within the company.

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
damien-urruty-sonarsource: I tried to find a common name for both approaches, v2 API for SQS and API subdomain for SQC. New APIs are exposed via v2 and subdomain, as it is the case for SCA, and I tried to group the management of those 2 things with the same method. Do you have a suggestion for a better name, or a better design? Can it be implemented as a later improvement?

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
damien-urruty-sonarsource: See my first answer about this point, it's not that simple

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
vnaskos-sonar: For the v2 I would suggest to revert the changes and add an if on the `ScaAPI` to use either the SQC endpoint or the SQS endpoint (Still not the best solution). It will be simpler to improve it later on.

src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
vnaskos-sonar: An idea would be 
* to remove the `SONARQUBE_CLOUD_URL` since we only need the api url.
* keep the `SONARQUBE_CLOUD_API_URL` for internal purposes.
* Introduce a `REGION` parameter.
* and just ask the organization and region from the users.

That way, the api url will always be correct, there won't be any discrepancy between the URL and the API_URL and we will fully support all SQC instances. However, moving those changes to another task/PR seems more clear, as they are a different topic than the SQC SCA support.

damien-urruty-sonarsource: I simplified the code to focus on SCA, and I created a separate ticket for that: https://sonarsource.atlassian.net/browse/MCP-144

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
damien-urruty-sonarsource: Will also be handled as part of https://sonarsource.atlassian.net/browse/MCP-144

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
damien-urruty-sonarsource: That's what I did

PR #109, opened 2025-10-06
MCP-156 Stop checking SCA enablement at startup

damien-urruty-sonarsource:
[MCP-156](https://sonarsource.atlassian.net/browse/MCP-156)

[MCP-156]: https://sonarsource.atlassian.net/browse/MCP-156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
@@ -114,16 +114,6 @@ public class SonarQubeMcpServer {
         new SystemStatusTool(serverApi)));
     }
 
-    if (mcpConfiguration.isSonarCloud() || sonarQubeVersionChecker.isSonarQubeServerVersionHigherOrEqualsThan("2025.4")) {
-      if (serverApi.scaApi().getFeatureEnabled().enabled()) {
-        this.supportedTools.add(new SearchDependencyRisksTool(serverApi));
-      } else {
-        LOG.info("Search Dependency Risks tool is not available because Advanced Security is not enabled.");
-      }
-    } else {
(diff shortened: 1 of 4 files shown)


nquinquenel (commented):
I personally believe that we should not enable tools that won't work. The protocol is made so that you can dynamically enable or disable tools at runtime. I think we should detect whether the configuration is incorrect and avoid enabling such tools.

damien-urruty-sonarsource:
Here are some thoughts:

* For this specific tool, SCA enablement can change over time. It might not be enabled at startup but it can be enabled later, e.g. from a browser. In this situation, the tool wouldn't be usable until MCP is restarted
* Skipping tools at startup only print something in the logs, and only for some tools, some others are just silently skipped. It's not very useful for the user/agent, both in terms of discovery (for them those tools simply don't exist), and in terms of remediation. If SCA is disabled, the agent could give advice on how to enable it, link to some relevant doc, etc. Even if the SQS instance is too old for SCA, the agent could also help and provide instructions to update. As a weaker argument, with time this condition will disappear as we will bump the minimal version
* if the tools are not available at all on a given platform, e.g. the system tools available only on SQS, I think the reasoning is a bit different, and for this I have more mixed feelings. Anyway I had created a separate [issue](https://sonarsource.atlassian.net/browse/MCP-154) for this.
* declaring all tools and then almost instantly send a notification to the client to disable them looks a bit hacky to me, this is probably subjective but I doubt it's a better solution. We also trust the clients to properly react to this notification, which should be OK in most cases I guess, but we would need to test it a bit with different clients
* my main concern at the moment is to fix the publication on the Docker MCP registry and unblock the situation. We could decide to go for a quick (and maybe imperfect) solution now, and revisit this choice later. If you have another idea to quickly unlock the situation I'd be happy to hear it

PR #220, opened 2026-02-15
MCP-295 Add detailed failure context to QG status tool

nquinquenel:


voz:
@nquinquenel can we close it or define path forward if we need to? Thx for helping to keep things tidy.

PR #335, opened 2026-04-13
MCP-417 Improve proxied server resilience

nquinquenel:


its/src/test/java/org/sonarsource/sonarqube/mcp/its/ProxiedServerITest.java
```diff
@@ -53,18 +58,39 @@ void should_successfully_connect_to_cag_proxied_server_when_given_enough_time()
       .contains("Successfully loaded 1 proxied MCP server(s)")
       .contains("Initializing 1 proxied MCP server(s)")
       .contains("Connecting to 'caas'")
-      .contains("Connected to 'caas' - discovered 1 tool(s)")
+      .containsPattern("Connected to 'caas' - discovered 10 tool\\(s\\)")
       .contains("MCP client manager initialization completed. 1/1 server(s) connected")
-      .contains("Loaded 1 proxied tool(s) from 1/1 server(s)");
+      .containsPattern("Loaded 10 proxied tool\\(s\\) from 1/1 server\\(s\\)")
+      .contains("All tools loaded: 34 tools after category filtering");
```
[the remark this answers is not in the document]
nquinquenel: It's expected

gradle.properties
```diff
@@ -1,5 +1,5 @@
 version=1.17-SNAPSHOT
-omnisharpVersion=1.39.10
```
sophio-japharidze-sonarsource, line 2: Just making sure omnisharp removal was intentional?

Dockerfile
```diff
@@ -43,6 +43,7 @@ RUN apk upgrade --no-cache && \
         chmod 0440 /etc/sudoers.d/appuser
 
 ARG TARGETARCH
+# Keep in sync with sonarContextAugmentationVersion in gradle.properties
```
sophio-japharidze-sonarsource, line 46: just out of curiosity, what happens if they are not in sync?

src/main/java/org/sonarsource/sonarqube/mcp/client/ProxiedToolsLoader.java
```diff
@@ -119,4 +135,22 @@ public void shutdown() {
     }
   }
 
+  static boolean isCommandAvailable(String command) {
+    if (new File(command).isAbsolute()) {
+      var file = new File(command);
+      return file.isFile() && file.canExecute();
+    }
```
sophio-japharidze-sonarsource: Nitpick 😄 
```suggestion
  var file = new File(command);
  if (file.isAbsolute()) {
    return file.isFile() && file.canExecute();
  }
```

sophio-japharidze-sonarsource (approved):

gradle.properties
nquinquenel: Yes, there was never any omnisharp in this repo, it was a mistake from the very beginning when we copied the existing property file from SLI, I find it funny that we notice it now 😅

Dockerfile
nquinquenel: Nothing - the version in `gradle.properties` is used for the ITs, we want to make sure that we test the current version shipped in the Dockerfile. I couldn't find a proper way to have both values centralized, because if you pass with via `ARG` it works from our CI, but Docker's registry doesn't and it wouldn't have the correct version.

PR #338, opened 2026-04-16
MCP-420 Remove redundant warning about bind address

sophio-japharidze-sonarsource:


.claude/settings.local.json
```diff
@@ -0,0 +1,12 @@
...
+    "allow": [
+      "mcp__atlassian__search",
+      "mcp__atlassian__getJiraIssue",
+      "mcp__atlassian__getJiraIssueRemoteIssueLinks",
+      "mcp__atlassian__getVisibleJiraProjects",
+      "mcp__atlassian__createJiraIssue",
+      "Bash(./gradlew test *)"
+    ]
+  }
+}
```
nquinquenel: Can you remove this please? 😅

src/main/java/org/sonarsource/sonarqube/mcp/transport/HttpServerTransportProvider.java
```diff
@@ -148,7 +148,7 @@ public HttpServerTransportProvider(int port, String host, AuthMode authMode, boo
       + protocol + "://" + host + ":" + port + MCP_ENDPOINT + " with authentication: " + authMode);
 
     // Warn about security risk when binding to all interfaces
-    if ("0.0.0.0".equals(host)) {
+    if ("0.0.0.0".equals(host) && !httpsEnabled) {
```
nquinquenel: Ideally, this warning should be removed only in a Docker environment, it's still valid if the MCP is run via a JAR

nquinquenel: Also it doesn't matter whether it's HTTP or HTTPS, this binding impacts the surface attack, TLS just encrypt the transport but doesn't reduce the surface.

nquinquenel (commented):
This warning should be kept if the MCP is running via a JAR, it's a big security risk (DNS rebinding). 0.0.0.0 is only necessary in Docker environment.

Also, we should probably mention something new in the README + Doc:
- If you are running the server locally on your machine, then add `-p 127.0.0.1:8443:8443`
- If it's meant to run remotely, then we can keep `-p 8443:8443`

src/main/java/org/sonarsource/sonarqube/mcp/transport/HttpServerTransportProvider.java
```diff
@@ -288,6 +291,10 @@ public String getServerUrl() {
     return protocol + "://" + host + ":" + port + MCP_ENDPOINT;
   }
 
+  private static boolean isRunningInDocker() {
+    return new File("/.dockerenv").exists();
+  }
```
nquinquenel: I think we could rely on a specific env var in the `Dockerfile`, like

`ENV SONARQUBE_MCP_IN_CONTAINER=true`

and in `McpLaunchConfiguration` class we could add a new variable `isRunningInContainer` that resolves this value, and use it inside this one.

Usually, we keep all configuration in this one `McpLaunchConfiguration` class

nquinquenel (commented):
Looks good, just the way to resolve whether we're running in a container could be changed!

nquinquenel (approved):
LGTM!

PR #350, opened 2026-04-21
MCP-430 Nudge agents toward Context Augmentation tools when CAG is enabled

AbbasNS:
## Summary
- attempt to replace md file documented in https://docs.sonarsource.com/sonarqube-cloud/analyzing-source-code/context-augmentation#integration-steps 
- When the organization has Context Augmentation entitled and the CAG toolset is active in stdio mode, append a section to the MCP server's composed instructions describing the 11 CAG tools and when to call each.
- The nudge is gated alongside the existing HTTP-mode and toolset checks at the call site, so non-CAG sessions (disabled toolset, HTTP transport, org without entitlement) are unaffected.
- Tested on Claude Code, Cursor, and Codex, and lead to significant call boost of CAG tools at the right time.


AbbasNS:
I believe the ITs failures are unrelated to my changes

kornel-kotan-sonarsource (approved):
LGTM, I have verified that the CAG Tools are being called more frequently with this change

AbbasNS:
Hi @nquinquenel, Context:
* We are trying to remove this https://docs.sonarsource.com/sonarqube-cloud/analyzing-source-code/context-augmentation#integration-steps from the recommended setup.
* not many users are setting it, it is manual, and get outdated, and the MCP instruction feature seems more reliable https://modelcontextprotocol.io/specification/draft/schema#initializeresult-instructions
* It is only injected when CAG is enabled
* I and Kornel tested it on Cursor, Claude Code, and Codex. The results looks satisfying.
* the IT failures seems unrelated to the changes and on other PRs.

 let me know if you would like more details around this

nquinquenel:
@AbbasNS There is already a dedicated `instructions` field in `src/main/resources/proxied-mcp-servers.json` that allows proxied servers like yours to declare additional instructions, which are then appended to ours.

If you take a look at the `JsonServerConfig` class, you’ll see the `instructions` field there (it’s currently nullable, so it does not appear in the JSON file, as we don't have any instructions yet).

Would that be sufficient? What do you think?

AbbasNS:
@nquinquenel, Good suggestion!
The only inconvenience, is that json is not very readable, and we might have some formatting issues unnoticed.
I added an equivalent md file to the test to mitigate this.

For Later, I will create a ticket on our side to provide the server instruction dynamically, and we can change the code here to do:
```
public String composeInstructions(String baseInstructions) {
  var builder = new StringBuilder(baseInstructions);
  mcpClientManager.getServerInstructions().values().stream()
      .filter(s -> s != null && !s.isBlank())
      .forEach(s -> builder.append("\n\n").append(s));
  return builder.toString();
}
```

nquinquenel:
@AbbasNS Sounds good, I think we can definitely improve this in the future, I agree that it's currently not ideal!

PR #353, opened 2026-04-23
docs: Add one-click "Add to Kiro" install badge

AnilMaktala:
Adds one-click "Add to Kiro" install badge to the existing Kiro section.

When clicked, Kiro auto-configures this MCP server — no manual JSON editing needed.

- Added "Add to Kiro" badge inside the existing Kiro `<details>` section (SonarQube Cloud config)
- Manual JSON config already present is preserved as fallback

Only the README is modified — no other content changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,8 @@ For **SonarQube Cloud US**, add `"-e", "SONARQUBE_URL=$SONAR_URL"` to the `args`
 <details>
 <summary>Kiro</summary>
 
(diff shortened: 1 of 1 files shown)


nquinquenel:
Hello @AnilMaktala, many thanks for your contribution. This is a valuable idea, and I tracked this effort in our backlog.

That said, we’re currently revisiting our contribution guidelines, and contributions from external users are becoming more restricted for the time being. Because of this, we’d prefer to implement the changes ourselves internally.

Would that be okay with you?

PR #367, opened 2026-04-27
MCP-443 Release version 1.17

nquinquenel:


gemini-extension.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sonarqube-mcp-server",
   "description": "The SonarQube MCP Server extension enables seamless integration with SonarQube Server or Cloud for code quality and security. It also supports the analysis of code snippet directly within the agent context.",
-  "version": "1.16",
+  "version": "1.17",
```
sophio-japharidze-sonarsource: Let's skip this change in the current PR & update it in #352 . Like that when current Gemini users detect a version update, they will get properly migrated to the new extension in the new repo.

sophio-japharidze-sonarsource (changes requested):

gemini-extension.json
nquinquenel: Done!

sophio-japharidze-sonarsource (approved):

PR #375, opened 2026-04-28
MCP-432 Forward mcp_server_id in _meta of InitializeRequest

nquinquenel:
Not ideal due to SDK limitation, issue on their side: https://github.com/modelcontextprotocol/java-sdk/issues/940

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/client/InitializeMetaInjectingClientTransport.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/client/InitializeMetaInjectingClientTransport.java
@@ -1,0 +1,105 @@
+/*
+ * SonarQube MCP Server
+ * Copyright (C) SonarSource
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonarsource.sonarqube.mcp.client;
+
+import io.modelcontextprotocol.json.TypeRef;
+import io.modelcontextprotocol.spec.McpClientTransport;
+import io.modelcontextprotocol.spec.McpSchema;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import reactor.core.publisher.Mono;
+
+/**
+ * Decorator around an {@link McpClientTransport} that rewrites outgoing {@code initialize} JSON-RPC requests so that
+ * their {@code _meta} field carries startup telemetry correlation data (e.g. {@code mcp_server_id},
+ * {@code invocation_id}).
+ *
+ * <p>The MCP Java SDK's {@code McpSyncClient#initialize()} does not expose a way to set {@code _meta} on the
+ * {@link McpSchema.InitializeRequest} it builds internally. By wrapping the underlying transport, we can intercept the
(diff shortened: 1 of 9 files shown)


nquinquenel:
@sonar-review-alpha review

src/main/java/org/sonarsource/sonarqube/mcp/client/InitializeMetaInjectingClientTransport.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.function.Consumer;
+import java.util.function.Function;
+import reactor.core.publisher.Mono;
+
+/**
+ * Decorator around an {@link McpClientTransport} that rewrites outgoing {@code initialize} JSON-RPC requests so that
+ * their {@code _meta} field carries startup telemetry correlation data (e.g. {@code mcp_server_id},
+ * {@code invocation_id}).
+ *
+ * <p>The MCP Java SDK's {@code McpSyncClient#initialize()} does not expose a way to set {@code _meta} on the
```
damien-urruty-sonarsource: Could you link the GH issue so we have a reference to the feature request close to the code?

src/main/java/org/sonarsource/sonarqube/mcp/client/InitializeMetaInjectingClientTransport.java
```diff
@@ -0,0 +1,105 @@
...
+ * {@code tools/call}. All other transport responsibilities are delegated unchanged.
+ */
+public class InitializeMetaInjectingClientTransport implements McpClientTransport {
+
+  private final McpClientTransport delegate;
+  private final Map<String, Object> initializeMeta;
+
+  public InitializeMetaInjectingClientTransport(McpClientTransport delegate, Map<String, Object> initializeMeta) {
+    this.delegate = delegate;
+    this.initializeMeta = Map.copyOf(initializeMeta);
```
damien-urruty-sonarsource: I remember recently fixing a bug with `Map.copyOf` when the map contains null values. I think we use this method in lots of places, do we need to do that? I think we could trust the map not to change after it's constructed in this case, it's unnecessary defensive

src/main/java/org/sonarsource/sonarqube/mcp/client/InitializeMetaInjectingClientTransport.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Override
+  public List<String> protocolVersions() {
+    return delegate.protocolVersions();
+  }
+
+  McpSchema.JSONRPCMessage maybeInjectInitializeMeta(McpSchema.JSONRPCMessage message) {
+    if (initializeMeta.isEmpty()) {
+      return message;
+    }
```
damien-urruty-sonarsource: Instead of doing this check for every message, I think we could do it once before instantiating `InitializeMetaInjectingClientTransport`. If it's empty, we can use the SDK transport, else we can use this class. 

EDIT: I see that the Map is always not empty, so we could also remove this check

src/main/java/org/sonarsource/sonarqube/mcp/client/McpClientManager.java
```diff
@@ -231,6 +236,13 @@ Map<String, String> buildEnvironmentVariables(ProxiedMcpServerConfig config, Map
     return filteredEnv;
   }
 
+  private McpClientTransport wrapWithInitializeMeta(McpClientTransport transport) {
+    var meta = Map.<String, Object>of(
+      "mcp_server_id", mcpServerId,
+      "invocation_id", UUID.randomUUID().toString());
```
damien-urruty-sonarsource: What does this invocation ID represent? We already pass an invocation ID in the meta of a tool call, it's a bit confusing if we use the same name here

damien-urruty-sonarsource (approved):
LGTM, some suggestions

src/main/java/org/sonarsource/sonarqube/mcp/client/InitializeMetaInjectingClientTransport.java
nquinquenel: Indeed, the MCP server ID should always be present. I'm removing this.

src/main/java/org/sonarsource/sonarqube/mcp/client/McpClientManager.java
nquinquenel: Arf, good catch, we don't need it anymore. It originated from the first implementation discussions but it was removed as it doesn't make sense (it has no use, and it's not really an invocation).

PR #376, opened 2026-04-28
CAG-296 Forward CAG instructions from proxied InitializeResult

AbbasNS:
## Summary

The CAG nudge text was previously bundled in this repo at `src/main/resources/proxied-mcp-servers.json` and appended via `ProxiedToolsLoader.composeInstructions` (introduced in #350). That coupling has three problems:

1. **Wrong owner.** The CAG squad owns the tools but cannot ship instruction changes without a `sonarqube-mcp-server` release.
2. **Drift.** CAG can rename / add / remove tools between releases; the bundled text goes stale.
3. **Unconditional appending.** The text is appended even when the CAG binary failed to launch, when SCA is disabled, or when architecture/guidelines kill switches are set on the CAG side. None of those runtime states are visible from this repo.

The MCP protocol already carries an `instructions: string` field in `InitializeResult`. Letting the CAG server own that field — and forwarding it from here — fixes all three problems and turns availability into a natural precondition: no successful `initialize` → no instructions appended.

This branch is rebased on top of [#378 (Pin CAG 0.8.0.355)](https://github.com/SonarSource/sonarqube-mcp-server/pull/378), the first CAG release that ships instructions in its `InitializeResult`. #378 must merge first.

## What changed

- `ProxiedToolsLoader.composeInstructions` now takes the runtime instructions list captured from each proxied server's `InitializeResult` instead of static config text.
- `McpClientManager` records `initializeResult.instructions()` per connected server and exposes them via `getProxiedInstructions()`.
- The `instructions` field is removed from `ProxiedMcpServerConfig` and the bundled `cag-instructions.md` is deleted.
- `SonarQubeMcpServer.loadProxiedServerTools` logs `Forwarded instructions from N proxied server(s)` for observability.


src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
```diff
@@ -452,10 +451,10 @@ private void loadProxiedServerTools() {
     var proxiedTools = proxiedToolsLoader.loadProxiedTools(currentTransportMode, mcpConfiguration.getMcpServerId());
     supportedTools.addAll(proxiedTools);
 
-    var parseResult = ProxiedServerConfigParser.parse();
-    if (parseResult.success() && !parseResult.configs().isEmpty()) {
-      composedInstructions = ProxiedToolsLoader.composeInstructions(composedInstructions, parseResult.configs());
-    }
+    var proxiedInstructions = proxiedToolsLoader.getProxiedInstructions();
+    composedInstructions = ProxiedToolsLoader.composeInstructions(composedInstructions, proxiedInstructions);
+    LOG.info("Forwarded instructions from " + proxiedInstructions.size() + " proxied server(s)");
```
nquinquenel: Let's log this in `debug`, it's something internal that we do not want to expose

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
```diff
@@ -452,10 +451,10 @@ private void loadProxiedServerTools() {
...
+    proxiedInstructions.forEach(s -> LOG.debug("Proxied instructions: " + s));
```
nquinquenel: I'd use parameterised form: `LOG.debug("Proxied instructions: {}", s)`

its/src/test/java/org/sonarsource/sonarqube/mcp/its/ProxiedServerITest.java
```diff
@@ -61,6 +61,9 @@ void should_successfully_connect_to_cag_proxied_server_when_given_enough_time()
       .containsPattern("Connected to 'caas' - discovered 10 tool\\(s\\)")
       .contains("MCP client manager initialization completed. 1/1 server(s) connected")
       .containsPattern("Loaded 10 proxied tool\\(s\\) from 1/1 server\\(s\\)")
+      .contains("Forwarded instructions from 1 proxied server(s)")
+      .contains("Proxied instructions: ")
+      .contains("## Context Augmentation")
```
nquinquenel, line 66: The previous test `cag_instructions_in_bundled_config_must_match_cag_instructions_md` gave us a review gate on what gets injected into users' LLM context, and this PR deletes it without a replacement.

This test asserts `contains("## Context Augmentation")` which is a start, but it's a single string check that will pass even if the instructions are severely degraded. I'd like to see the IT assert the structural shape (if not the full content) of what CAG actually returns so that future CAG releases that change instructions must also update this file, giving us a review checkpoint of what is injected into users' context.

nquinquenel (commented):
Looks good, I'm mostly concerned about visibility on what we ship

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
AbbasNS: Fixed in 9b2b254

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
AbbasNS: fixed in 9b2b254

its/src/test/java/org/sonarsource/sonarqube/mcp/its/ProxiedServerITest.java
AbbasNS: >The previous test cag_instructions_in_bundled_config_must_match_cag_instructions_md gave us a review gate on what gets injected into users' LLM context, and this PR deletes it without a replacement.

The replacement tests are on the CAG side. Previously, the exact content was tested here because it was implemented here.
Is the goal on the SQ MCP side to assert the exact wording of the instructions, or simply to verify that the instructions are being communicated? 
I’m fine with either approach, but I’m not sure there’s value in failing the SQ MCP integration test every time there’s a wording change on the Context MCP.
For example, we assert that 10 proxied tools are loaded, but we don’t assert their names or descriptions.

Up to you if would like visibility on the content of instructions after every CAG release.

nquinquenel: I agree that we also don't test the tools you provide - maybe that's something we should do?

I'm fine leaving it like this for now, but we own the delivery and what we ship to users. I believe that having a clear understanding of what we ship (on our side) is important. I'll think about this. Given that everything is internal, and that I can also check your codebase, LGTM for now.

nquinquenel (approved):

PR #386, opened 2026-05-06
Redirect certificate installation logs to stderr in Docker stdio mode

angelique-lebail:
Title: Redirect certificate installation logs to stderr in Docker stdio mode

## Summary

This change redirects certificate installation informational messages from `stdout` to `stderr` in `scripts/install-certificates.sh`.

## Why

The Docker image is commonly used as an MCP stdio server.
In stdio mode, `stdout` must remain reserved for JSON-RPC protocol traffic.

When custom certificates are mounted into `/usr/local/share/ca-certificates/`, the current script writes:

- `Installing custom certificates from ...`
- `Custom certificates installed successfully`

to `stdout` before the Java MCP server starts.

That can break MCP clients during handshake.

## Change

- redirect the two informational `echo` calls to `stderr`

## Effect

- no behavior change for certificate installation itself
- logs remain visible
- `stdout` stays clean for MCP stdio clients

## Validation

Reproduced with a container run that mounts custom certificates into `/usr/local/share/ca-certificates/`.

Before this patch:

- informational certificate messages appeared on `stdout`

After this patch:

- `stdout` stays empty before MCP traffic
- certificate and warning messages are emitted on `stderr`

### DIFF
--- a/scripts/install-certificates.sh
+++ b/scripts/install-certificates.sh
@@ -8,7 +8,7 @@ if [ "$(ls -A "$CERT_DIR")" ]; then
     echo "Installing custom certificates from $CERT_DIR..." >&2
 
     # Run as root via sudo
-    sudo /usr/sbin/update-ca-certificates
+    sudo /usr/sbin/update-ca-certificates >&2
 
     echo "Custom certificates installed successfully" >&2
 fi
--- a/scripts/install-certificates.sh
+++ b/scripts/install-certificates.sh
@@ -5,10 +5,10 @@
 CERT_DIR="/usr/local/share/ca-certificates/"
 
 if [ "$(ls -A "$CERT_DIR")" ]; then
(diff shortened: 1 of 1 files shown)


nquinquenel:
Hello @angelique-lebail, thank you very much for your contribution! I think your concern is valid, and I’d like to move forward with it.

That said, we’re currently revisiting our contribution guidelines, and contributions from external users are becoming more restricted for the time being. Because of this, we’d prefer to implement the changes ourselves internally.

Would that be okay with you?

angelique-lebail:
Hello @nquinquenel, thanks for the update and for considering the contribution.
That’s totally okay with me — I understand the situation. I’m glad the concern was considered useful, and I appreciate your feedback.

nquinquenel:
Thank you @angelique-lebail, this effort is now tracked in our backlog, and we will try to implement it in an upcoming release!

PR #387, opened 2026-05-06
Fix missing @Nullable on optional Issue fields causing outputSchema validation errors

devjhan:
* Please review our [contribution guidelines](https://github.com/SonarSource/sonarqube-mcp-server/blob/master/docs/contributing.md).

And please ensure your pull request adheres to the following guidelines:

- [x] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [x] Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [x] Provide a unit test for any code you changed
- [ ] If there is a [JIRA](http://jira.sonarsource.com/browse/MCP) ticket available, please make your commits and pull request start with the ticket ID (MCP-XXXX)

---

## Problem

`search_sonar_issues_in_projects` returns a `tools/call` validation error whenever the SonarQube response contains an issue where `author` (or other optional fields) is absent. This makes the tool effectively unusable for any project that includes such issues, and there is no argument-level workaround available to the user.

The root cause is a mismatch between two mechanisms in `SchemaUtils`:

1. `withRequiredCheck` marks every field without `@Nullable` as `required` in the generated `outputSchema`.
2. `ObjectMapper` is configured with `JsonInclude.Include.NON_NULL`, which silently omits null fields from the serialized payload.

When SonarQube legitimately omits a field (e.g. `author` when SCM blame has not been performed), the Java field is set to `null` by Gson, the serializer drops the key, and the resulting payload fails the `required` constraint in the MCP `outputSchema`. Because this can affect any issue in the response, filtering or pagination arguments cannot work around it.

## Fix

Added `@Nullable` to the four fields in `SearchIssuesToolResponse.Issue` that SonarQube can legitimately omit:

| Field | When absent |
|---|---|
| `author` | SCM blame not run, external/imported issue, or user anonymization (SonarQube Cloud) |
| `severity` | MQR (Multi-Quality Rating) mode — legacy single severity is replaced by `impacts[]` |
| `cleanCodeAttribute` | Legacy rules or older SonarQube Server versions |
| `cleanCodeAttributeCategory` | Same as `cleanCodeAttribute` |

With `@Nullable` in place, `SchemaUtils.withRequiredCheck` excludes these fields from the `required` array in the output schema, so a missing field no longer causes a validation failure. The fields remain in `properties`, so clients that do receive them can still consume them normally.

## Tests

- Updated `it_should_validate_output_schema_and_annotations` to reflect the corrected `required` array (the four fields above removed).
- Added `it_should_handle_issues_with_null_author_and_other_optional_fields` to directly reproduce the reported failure: a WireMock-stubbed SonarQube response with `author`, `severity`, `cleanCodeAttribute`, and `cleanCodeAttributeCategory` all absent is returned successfully without a validation error.

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/tools/issues/SearchIssuesToolResponse.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/tools/issues/SearchIssuesToolResponse.java
@@ -35,12 +35,12 @@ public record SearchIssuesToolResponse(
     @JsonPropertyDescription("Rule that triggered the issue") String rule,
     @JsonPropertyDescription("Project key where the issue was found") String project,
     @JsonPropertyDescription("Component (file) where the issue is located") String component,
-    @JsonPropertyDescription("Issue severity level (absent in MQR mode)") @Nullable String severity,
+    @JsonPropertyDescription("Issue severity level") @Nullable String severity,
     @JsonPropertyDescription("Current status of the issue") String status,
     @JsonPropertyDescription("Issue description message") String message,
-    @JsonPropertyDescription("Clean code attribute associated with the issue (absent on legacy rules or older SonarQube Server)") @Nullable String cleanCodeAttribute,
-    @JsonPropertyDescription("Clean code attribute category (absent on legacy rules or older SonarQube Server)") @Nullable String cleanCodeAttributeCategory,
-    @JsonPropertyDescription("Author who introduced the issue (absent when SCM blame unavailable, or anonymized)") @Nullable String author,
+    @JsonPropertyDescription("Clean code attribute associated with the issue") @Nullable String cleanCodeAttribute,
+    @JsonPropertyDescription("Clean code attribute category") @Nullable String cleanCodeAttributeCategory,
+    @JsonPropertyDescription("Author who introduced the issue") @Nullable String author,
     @JsonPropertyDescription("Date when the issue was created") String creationDate,
     @JsonPropertyDescription("Location of the issue in the source file") @Nullable TextRange textRange
   ) {}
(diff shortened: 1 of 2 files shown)


nquinquenel:
Hello @devjhan, thank you for your contribution! I believe the concern you raised is valid.

At the moment, though, we’re in the process of updating our contribution guidelines, and external contributions are being handled more selectively. For that reason, we’d prefer to take care of the implementation on our side.

Would that be ok for you?

devjhan:
> Hello @devjhan, thank you for your contribution! I believe the concern you raised is valid.
> 
> At the moment, though, we’re in the process of updating our contribution guidelines, and external contributions are being handled more selectively. For that reason, we’d prefer to take care of the implementation on our side.
> 
> Would that be ok for you?

Totally makes sense, thanks for being upfront about it! Happy to close this out whenever you're ready.

Side note — if it's not too much trouble, a `Co-authored-by` credit on the commit would be a nice touch if the approach ends up being similar. But absolutely no worries if that's outside your usual process.

Either way, glad the issue is on your radar!

nquinquenel:
> > Hello @devjhan, thank you for your contribution! I believe the concern you raised is valid.
> > At the moment, though, we’re in the process of updating our contribution guidelines, and external contributions are being handled more selectively. For that reason, we’d prefer to take care of the implementation on our side.
> > Would that be ok for you?
> 
> Totally makes sense, thanks for being upfront about it! Happy to close this out whenever you're ready.
> 
> Side note — if it's not too much trouble, a `Co-authored-by` credit on the commit would be a nice touch if the approach ends up being similar. But absolutely no worries if that's outside your usual process.
> 
> Either way, glad the issue is on your radar!

Unfortunately, we do not let users sign CLA for the time being, which means we cannot accept your contribution as a `Co-authored-by`. We will soon proceed to open the `Issues` section on GitHub, which will let users report issues without having to suggest code. Sorry for the inconvenience.

PR #397, opened 2026-05-15
SONAR-27552: Add Docker publish workflow for MCP server

carminevassallo:
## SONAR-27552: Add Docker publish workflow for MCP server

Introduces a CI workflow to build and publish the MCP server Docker image to [sonarsource/sonarqube-mcp](https://hub.docker.com/r/sonarsource/sonarqube-mcp) on Docker Hub.

## Changes

**`Dockerfile`**
- Parameterized the JAR download URL via `ARG MCP_SERVER_VERSION` (default `1.18.1.2664` preserved for local builds and PR dry runs).

**`.github/workflows/docker-publish.yml`** (new)
- `release` trigger: builds and pushes `<version>` + `latest` tags automatically when a release is published.
- `workflow_dispatch` trigger: on-demand publish with a required `tag` input; checks out the exact git tag snapshot. A `dry_run` boolean (default: `true`) allows build validation without pushing.
- `pull_request` trigger: dry-run only (never pushes), checks out the PR head SHA, and reads the version from the Dockerfile `ARG` default.
- Multi-platform build (linux/amd64 + linux/arm64) via QEMU and Docker Buildx with the `docker-container` driver.
- Smoke test: starts the container with `SONARQUBE_IS_CLOUD=true` and waits for `SonarQube MCP Server Started` in the logs; runs on each platform's native runner using a shared temp image from [ttl.sh](https://ttl.sh).
- Docker Hub credentials retrieved from Vault (`development/kv/data/docker/sonardockerrw`).
- User-controlled inputs (`inputs.tag`, `inputs.dry_run`) are assigned to environment variables before being used in shell scripts to prevent script injection.

## Workflow design

```
prepare  →  build (multi-platform → ttl.sh)  →  test (amd64 + arm64, native runners)  →  publish (rebuild → Docker Hub)
```

The publish job rebuilds the image instead of promoting the temp image to avoid any provenance metadata referencing the ttl.sh registry.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -60,9 +60,9 @@ jobs:
             VERSION=$(grep 'ARG MCP_SERVER_VERSION=' Dockerfile | cut -d= -f2)
           fi
 
-          TAGS="sonarsource/sonarqube-mcp-server:${VERSION}"
+          TAGS="sonarsource/sonarqube-mcp:${VERSION}"
           if [ "${{ github.event_name }}" = "release" ]; then
-            TAGS="${TAGS}"$'\n'"sonarsource/sonarqube-mcp-server:latest"
+            TAGS="${TAGS}"$'\n'"sonarsource/sonarqube-mcp:latest"
           fi
 
           if [ "${{ github.event_name }}" = "release" ] || [ "${INPUT_DRY_RUN}" = "false" ]; then
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -82,26 +82,10 @@ jobs:
         with:
           ref: ${{ needs.prepare.outputs.ref }}
 
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
-        with:
-          name: multibuilder
-          driver: docker-container
-          install: true
-
-      - name: Set up QEMU for multi-arch builds
-        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
-        with:
-          platforms: all
-
       - name: Build and push to temporary registry
-        uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f # v7.1.0
+        uses: ./.github/actions/docker-build
         with:
-          context: .
-          platforms: linux/amd64,linux/arm64
-          push: true
-          build-args: |
-            MCP_SERVER_VERSION=${{ needs.prepare.outputs.version }}
+          version: ${{ needs.prepare.outputs.version }}
           tags: ${{ needs.prepare.outputs.temp_image }}
 
   test:
@@ -116,6 +100,7 @@ jobs:
           - platform: linux/arm64
             runner: github-ubuntu-24.04-arm-s
     runs-on: ${{ matrix.runner }}
+    permissions: {}
     steps:
       - name: Smoke test
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,99 @@
...
+          ref: ${{
+            github.event_name == 'workflow_dispatch' && format('refs/tags/{0}', inputs.tag) ||
+            github.event_name == 'pull_request' && github.event.pull_request.head.sha ||
+            '' }}
+
+      - name: Resolve version
+        id: version
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            VERSION="${{ inputs.tag }}"
```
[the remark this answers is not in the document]
carminevassallo: Fixed.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,187 @@
...
+      version: ${{ steps.version.outputs.value }}
+      tags: ${{ steps.version.outputs.tags }}
+      push: ${{ steps.version.outputs.push }}
+      ref: ${{ steps.ref.outputs.value }}
+    steps:
+      - name: Compute checkout ref
+        id: ref
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "value=refs/tags/${{ inputs.tag }}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
carminevassallo: Fixed.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,213 @@
...
+      - name: Compute build parameters
+        id: params
+        env:
+          INPUT_TAG: ${{ inputs.tag }}
+          INPUT_DRY_RUN: ${{ inputs.dry_run }}
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            VERSION="${INPUT_TAG}"
+          elif [ "${{ github.event_name }}" = "release" ]; then
+            TAG="${{ github.event.release.tag_name }}"
```
nquinquenel: `github.event.release.tag_name` interpolated directly into shell, I think it should go through an env var like inputs already do (tag or dry run)

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,213 @@
+name: Publish Docker Image
+
+on:
+  pull_request:
+  release:
+    types: [published]
```
nquinquenel: I think this should be removed. Our release flow is:
- Release -> generate new JAR and publish it to binaries
- Open PR -> update Dockerfile with new version for the JAR

If we trigger this flow on release, it will still point to the previous version. It has to be done manually afterwards

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,213 @@
+name: Publish Docker Image
+
+on:
+  pull_request:
```
nquinquenel: Not sure I get this one, I think we should remove it no?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,213 @@
...
+        uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f # v7.1.0
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          build-args: |
+            MCP_SERVER_VERSION=${{ needs.prepare.outputs.version }}
+          tags: ${{ needs.prepare.outputs.temp_image }}
+
+  test:
```
nquinquenel, line 81: Should we add `permissions: {}` here, as best practice?

.github/workflows/docker-publish.yml
carminevassallo: Just left it for testing on this PR

.github/workflows/docker-publish.yml
carminevassallo: Ok we make the workflow only manually triggered

nquinquenel (approved):
LGTM

PR #406, opened 2026-05-18
MCP-481 Have GetComponentMeasuresTool use configured project key when available

sophio-japharidze-sonarsource:


nquinquenel (changes requested):
We should update the README as well, and ping doc team about this

nquinquenel (approved):

PR #413, opened 2026-05-22
Treat blank optional string arguments as absent

mysticfall:
Make getOptionalString return null for blank strings so that optional parameters with empty/whitespace-only values are treated the same as unset parameters.

This fixes an issue where invoking tools such as `search_security_hotspots` on a local project without a pull request ID fails, because the SonarQube Server's endpoint rejects a URL with an empty `pullRequest` parameter.

It's reproducible with OpenCode, using either a Docker or jar version of the project.

Part of 

* Please review our [contribution guidelines](https://github.com/SonarSource/sonarqube-mcp-server/blob/master/docs/contributing.md).

And please ensure your pull request adheres to the following guidelines: 

- [x] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [x] Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [x] Provide a unit test for any code you changed
- [ ] If there is a [JIRA](http://jira.sonarsource.com/browse/MCP) ticket available, please make your commits and pull request start with the ticket ID (MCP-XXXX)

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/tools/Tool.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/tools/Tool.java
@@ -134,7 +134,7 @@ public abstract class Tool {
     public String getOptionalString(String argumentName) {
       var arg = argumentsMap.get(argumentName);
       if (arg instanceof String string) {
-        return string;
+        return string.isBlank() ? null : string;
       } else {
         return null;
       }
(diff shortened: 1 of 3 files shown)


nquinquenel:
Hi @mysticfall, thanks for the PR and for the clear write-up and tests.

We’ve reviewed the change and agree with the diagnosis: when an MCP client sends an optional string as `""` or whitespace (e.g. `pullRequest` on a local project), we currently forward it as a query parameter and SonarQube Server rejects the request. Treating blank optional strings as absent in `getOptionalString()` is the right fix.

This repository currently doesn’t accept external pull requests anymore ([contributing guidelines](https://github.com/SonarSource/sonarqube-mcp-server/blob/master/docs/contributing.md)); we track work internally and land changes through our usual process. Your contribution is still valid - we’d like to implement a similar fix on our side and ship it in a follow-up release.

Thanks again for taking the time to report and fix this.

PR #417, opened 2026-05-27
Fix analyze_code_snippet silently returning 0 issues for .tsx/.jsx files

### LINKED ISSUE #396: BUG: `analyze_code_snippet` silently returns 0 issues on `.tsx`/`.jsx` (no `tsx`/`jsx` in `language` enum)
### `analyze_code_snippet` silently returns 0 issues on `.tsx`/`.jsx` (no `tsx`/`jsx` in `language` enum)

**Tool:** `analyze_code_snippet` · **Image:** `mcp/sonarqube:latest` · **CLI:** `v0.12.0`

#### Bug

The `language` enum exposes `ts` and `js` but no `tsx`/`jsx`. Passing `language: ["ts"]` on JSX-containing content makes the TS analyzer bail without error and return `{"issues":[],"issueCount":0}`. CI (`sonar-scanner`) on the same project + same quality profile catches the issues correctly — so this is specific to the MCP snippet path.

#### Reproducer

Pure TS — works:
```ts
// language: ["ts"]
export const f = (a?: string, b?: boolean) => a ? (b ? 'x' : 'y') : 'z';
// → typescript:S3358 ✅
```

Same logic in TSX — silent zero:
```tsx
// language: ["ts"]
import React from 'react';
const C = (p: { a?: string; b: boolean }) =>
  <div>{p.a ? (p.b ? 'x' : 'y') : 'z'}</div>;
export default C;
// → {"issues":[],"issueCount":0} ❌  (expected: typescript:S3358)
```

#### Workaround is unreliable

`language: ["js"]` parses JSX and surfaces `javascript:S3358` correctly, but the JS parser misreads TS generics — e.g. `React.useRef<HTMLDivElement | null>(null)` triggers a false `javascript:S1764` ("identical sub-expressions on both sides of `>`"). Also: `language: ["ts","js"]` returns 0 issues (possible separate bug).

#### Requested fix (any of)

1. Add `tsx`/`jsx` to the `language` enum and route to the TS/JS analyzers with JSX parsing on.
2. Have `["ts"]` accept TSX (TypeScript parses TSX natively given the right flag).
3. **At minimum:** return an error/diagnostic when the analyzer can't parse the snippet. The silent `{"issues":[],"issueCount":0}` is the dangerous part — agents trust it and report the file as clean.

#### Impact

Every `.tsx`/`.jsx` file in React/Next/etc. codebases gets a false-clean result. High blast radius for agentic use.

rishichintala:
Fixes #396

## Problem

`analyze_code_snippet` silently returns `{"issues":[],"issueCount":0}` when analyzing TSX/JSX content. The root cause is a two-step failure:

1. The `language` enum only exposes `ts` and `js` — there is no `tsx` or `jsx`.
2. When a user passes `language: "ts"` for TSX content, `createTemporaryFileForLanguage` creates a temp file with a `.ts` extension (derived from `SonarLanguage.TS.getDefaultFileSuffixes()`). The TypeScript analyzer **only enables JSX parsing for `.tsx` files** — for `.ts` files it silently bails on JSX syntax and reports zero issues, with no error or diagnostic.

Every `.tsx`/`.jsx` file in React/Next.js codebases gets a false-clean result. Agents trust it and report the file as clean.

## Fix

- Add `tsx` and `jsx` as accepted language aliases in `LanguageUtils`, mapping:
  - `tsx` → `SonarLanguage.TS` + `.tsx` temp file extension
  - `jsx` → `SonarLanguage.JS` + `.jsx` temp file extension
- When the user passes `language: "tsx"` or `language: "jsx"`, the temp file is created with the correct extension so the TS/JS analyzer enables JSX parsing
- The existing `ts` and `js` behaviour is unchanged

## Checklist

- [x] Explain motives: bug causes false-clean results on all TSX/JSX files in agentic workflows (described above and in #396)
- [x] Code style follows [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [x] Unit tests provided for all changed code (12 new tests in `LanguageUtilsTests`, including alias resolution, case-insensitivity, no-duplicate, extension mapping, and regression tests for existing `ts`/`js` behaviour)
- [x] Integration tests provided (`it_should_accept_tsx_as_a_valid_language`, `it_should_accept_jsx_as_a_valid_language` — server logs confirm `.tsx`/`.jsx` temp files are created)
- [ ] No JIRA ticket available (referencing GitHub issue #396)

### DIFF
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -53,7 +53,7 @@ The user should provide the following environment variables:
 - Python (including IPython notebooks)
 - Ruby
 - Go
-- JavaScript, TypeScript, JSP
+- JavaScript (js, jsx), TypeScript (ts, tsx), JSP
 - PHP
 - XML
 - HTML, CSS
--- a/README.md
+++ b/README.md
@@ -901,10 +901,10 @@ SOCKS5 proxies are supported.
   - `filePath` - Project-relative path of the file to analyze (e.g., `src/main/java/MyClass.java`). Used when the workspace is mounted at `/app/mcp-workspace` - _String_
   - `fileContent` - Complete file content as a string. Required when workspace is not mounted - _String_
   - `codeSnippet` - Code snippet to filter issues (must match content in fileContent) - _String_
-  - `language` - Language of the code (e.g., 'java', 'python', 'javascript') - _String_
+  - `language` - Language of the code (e.g., 'java', 'python', 'js', 'ts', 'tsx', 'jsx') - _String_
   - `scope` - Scope of the file: MAIN or TEST (default: MAIN) - _String_
   
-  **Supported Languages:** Java, Kotlin, Python, Ruby, Go, JavaScript, TypeScript, JSP, PHP, XML, HTML, CSS, CloudFormation, Kubernetes, Terraform, Azure Resource Manager, Ansible, Docker, Secrets detection
+  **Supported Languages:** Java, Kotlin, Python, Ruby, Go, JavaScript (`js`, `jsx`), TypeScript (`ts`, `tsx`), JSP, PHP, XML, HTML, CSS, CloudFormation, Kubernetes, Terraform, Azure Resource Manager, Ansible, Docker, Secrets detection
 
 **When integration with SonarQube for IDE is enabled:**
 - **analyze_file_list** - Analyze files in the current working directory using SonarQube for IDE. This tool connects to a running SonarQube for IDE instance to perform code quality analysis on a list of files.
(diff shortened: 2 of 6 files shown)


nquinquenel:
Hello @rishichintala , thank you for reporting this issue and suggesting a fix!

We’re currently revisiting our contribution guidelines, and contributions from external users are becoming more restricted for the time being. Because of this, we’d prefer to implement the changes ourselves internally.

Would that be okay with you?

rishichintala:
Hi @nquinquenel, Sure, I understand. Thanks for explaining.

I was hoping to contribute the fix myself since the project is open source,
but I understand the current restrictions around external contributions.

I’m glad my report and suggested approach were helpful. Please feel free to
proceed with implementing it internally, and let me know if you need any
more context from my side.

nquinquenel:
Thanks again for the clear write-up, your analysis and tests were a real help — appreciate the contribution.

PR #429, opened 2026-06-03
MCP-57 Expand on integration tests against SQC

nquinquenel:
----
## Summary by Gitar

- **Integration testing framework:**
  - Added `sonarCloudIntegrationTest` task and `SonarCloudStagingHarness` for in-process MCP client testing against real SonarQube Cloud staging.
  - Implemented a shared project provisioning mechanism using a `sample-java` Maven project to ensure consistent analysis state.
- **Coverage and verification:**
  - Added comprehensive tool-specific integration tests (`*SonarCloudIT`) to validate staging parity for every MCP tool.
  - Updated `ProxiedServerITest` to reflect increased tool count (34 to 40).
- **CI and infrastructure:**
  - Added `sonarcloud-integration.yml` weekly workflow and updated `build.yml` to run staging tests in parallel via matrix.
  - Configured failure notifications to include the new SonarQube Cloud integration suite.
- **Fixes:**
  - Corrected byte handling in `BlockingQueueOutputStream` (`b & 0xFF`) to prevent data corruption.


.github/workflows/build.yml
```diff
@@ -74,11 +79,16 @@ jobs:
             development/artifactory/token/{REPO_OWNER_NAME_DASH}-private-reader username | ARTIFACTORY_USER;
             development/artifactory/token/{REPO_OWNER_NAME_DASH}-private-reader access_token | ARTIFACTORY_ACCESS_TOKEN;
 
+      - name: Install Maven
+        if: matrix.sonarcloud == true
+        run: sudo apt-get update && sudo apt-get install -y maven
```
damien-urruty-sonarsource: Shouldn't we use Mise instead?

.github/workflows/sonarcloud-integration.yml
```diff
@@ -0,0 +1,43 @@
...
+    name: SonarQube Cloud staging (ITS)
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
+        with:
+          version: 2026.4.20
+
+      - name: Install Maven
+        run: sudo apt-get update && sudo apt-get install -y maven
```
damien-urruty-sonarsource: Same

.github/workflows/build.yml
```diff
@@ -74,11 +79,16 @@ jobs:
...
+
       - name: Download plugin from Artifactory
+        if: matrix.sonarcloud != true
         env:
-          ARTIFACTORY_URL: ${{ fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_URL }}
-          ARTIFACTORY_PRIVATE_USERNAME: ${{ fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_USER }}
-          ARTIFACTORY_PRIVATE_PASSWORD: ${{ fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_ACCESS_TOKEN }}
+          ARTIFACTORY_URL: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_URL || '' }}
+          ARTIFACTORY_PRIVATE_USERNAME: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_USER || '' }}
+          ARTIFACTORY_PRIVATE_PASSWORD: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_ACCESS_TOKEN || '' }}
```
damien-urruty-sonarsource: This syntax looks different than what we usually have, why do we need the `|| ''` at the end ? We should always expect the variable to be present

.github/workflows/build.yml
```diff
@@ -92,16 +102,31 @@ jobs:
...
+        if: matrix.sonarcloud == true
+        env:
+          SONARCLOUD_IT_TOKEN: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).SONARCLOUD_IT_TOKEN || '' }}
+          ARTIFACTORY_URL: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_URL || '' }}
+          ARTIFACTORY_USER: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_USER || '' }}
+          ARTIFACTORY_ACCESS_TOKEN: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_ACCESS_TOKEN || '' }}
+        run: |
+          if [ -z "${SONARCLOUD_IT_TOKEN}" ]; then
+            echo "::warning::SONARCLOUD_IT_TOKEN not available (e.g. fork PR); skipping SonarQube Cloud staging ITS"
+            exit 0
```
damien-urruty-sonarsource: We should fail harder than that IMO. It's somehow related to my previous comment, I don't know why we fallback to empty string in some cases

EDIT: I see your point about forks, but I think we shouldn't care too much about that

.github/workflows/sonarcloud-integration.yml
```diff
@@ -0,0 +1,43 @@
+name: SonarQube Cloud integration tests
+on:
+  schedule:
+    # Weekly on Monday at 06:00 UTC
+    - cron: '0 6 * * 1'
```
damien-urruty-sonarsource, line 5: I assume this is because there is low activity in the project?

.github/workflows/build.yml
```diff
@@ -65,38 +70,67 @@ jobs:
...
 
       - uses: SonarSource/ci-github-actions/config-gradle@de29c54be07c772c6afc381db8ae4bfcbfe4df12 # 1.8.0
         with:
           artifactory-reader-role: private-reader
 
+      - name: Install Maven
+        if: matrix.sonarcloud == true
+        run: sudo apt-get update && sudo apt-get install -y maven
+
       - name: Download plugin from Artifactory
```
damien-urruty-sonarsource: Not related to your changes but small typo here, should be binary or MCP JAR

.github/workflows/build.yml
```diff
@@ -65,38 +70,67 @@ jobs:
...
+          SONARCLOUD_IT_TOKEN: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).SONARCLOUD_IT_TOKEN || '' }}
+          ARTIFACTORY_URL: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_URL || '' }}
+          ARTIFACTORY_USER: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_USER || '' }}
+          ARTIFACTORY_ACCESS_TOKEN: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_ACCESS_TOKEN || '' }}
           DOWNLOADED_JAR_PATH: ${{ github.workspace }}/downloaded-plugin/sonarqube-mcp-server.jar
         run: |
           ./gradlew :its:integrationTest --tests "*${{ matrix.test_class }}" --info
 
+      - name: Run SonarQube Cloud staging integration tests
+        if: matrix.sonarcloud == true
```
damien-urruty-sonarsource: I wonder if relying on a Matrix brings a lot here, given we almost have 2 different branches. Maybe we could extract the SQC ITs in a separate job, or script? This way we could also factorize the script between the regular build and the weekly build?

its/projects/sample-java/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.sonarsource.sonarqube.mcp</groupId>
+  <artifactId>sample-java</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <name>MCP SonarCloud IT sample (Java)</name>
+  <description>Minimal Java project pushed to SonarQube Cloud staging for integration tests</description>
+
+  <properties>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
```
damien-urruty-sonarsource: Shouldn't we target 21?

its/src/test/java/org/sonarsource/sonarqube/mcp/its/sonarcloud/tools/SearchDependencyRisksSonarCloudIT.java
```diff
@@ -0,0 +1,48 @@
...
+import org.sonarsource.sonarqube.mcp.tools.dependencyrisks.SearchDependencyRisksTool;
+
+import static org.sonarsource.sonarqube.mcp.its.sonarcloud.harness.SonarQubeMcpTestClient.assertResultEquals;
+
+@Tag("SonarCloud")
+class SearchDependencyRisksSonarCloudIT extends AbstractSonarCloudStagingIT {
+
+  @Test
+  void should_call_search_dependency_risks_against_staging() {
+    assumeToolRegistered(SearchDependencyRisksTool.TOOL_NAME);
```
damien-urruty-sonarsource: Shouldn't we be more intentional, and fail the test if it's not the case? The problem I see is we could end up skipping the test and hiding a real problem

its/src/test/java/org/sonarsource/sonarqube/mcp/its/sonarcloud/AbstractSonarCloudStagingIT.java
```diff
@@ -0,0 +1,78 @@
...
+import org.sonarsource.sonarqube.mcp.its.sonarcloud.harness.SonarCloudAnalyzedProject;
+import org.sonarsource.sonarqube.mcp.its.sonarcloud.harness.SonarCloudStagingFixture;
+import org.sonarsource.sonarqube.mcp.its.sonarcloud.harness.SonarCloudStagingHarness;
+import org.sonarsource.sonarqube.mcp.its.sonarcloud.harness.SonarQubeMcpTestClient;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Tag("SonarCloud")
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public abstract class AbstractSonarCloudStagingIT {
```
damien-urruty-sonarsource, line 38: I often find tests depending on an abstract class harder to follow. Here we store little state in the abstract class, so it's probably OK. Another idea would be to move things under the harness and have it injected as a test method parameter, as we do in SLCORE. But tests would be slightly more verbose

its/src/test/java/org/sonarsource/sonarqube/mcp/its/ProxiedServerITest.java
```diff
@@ -58,13 +58,14 @@ void should_successfully_connect_to_cag_proxied_server_when_given_enough_time()
...
+      .containsPattern("Connected to 'caas' - discovered 11 tool\\(s\\)")
       .contains("MCP client manager initialization completed. 1/1 server(s) connected")
-      .containsPattern("Loaded 10 proxied tool\\(s\\) from 1/1 server\\(s\\)")
+      .containsPattern("Loaded 11 proxied tool\\(s\\) from 1/1 server\\(s\\)")
       .contains("Forwarded instructions from 1 proxied server(s)")
       .contains("Proxied instructions: ")
       .contains("## Context Augmentation")
-      .contains("All tools loaded: 34 tools after category filtering");
+      .contains("Advanced analysis mode enabled")
+      .contains("All tools loaded: 40 tools after category filtering");
```
damien-urruty-sonarsource, line 68: Do you know why we have a different number of tools loaded? It seems strange to have different expected things for a PR that only adds tests

its/src/test/java/org/sonarsource/sonarqube/mcp/its/sonarcloud/tools/ProjectQualityGateSonarCloudIT.java
```diff
@@ -0,0 +1,38 @@
...
+
+@Tag("SonarCloud")
+class ProjectQualityGateSonarCloudIT extends AbstractSonarCloudStagingIT {
+
+  @Test
+  void should_call_get_project_quality_gate_status_against_staging() {
+    var result = mcpClient.callTool(ProjectStatusTool.TOOL_NAME, Map.of("projectKey", fixture.projectKey()));
+
+    var content = structuredContent(result);
+    assertThat(content.get("status")).isIn("OK", "WARN", "ERROR", "NONE");
```
damien-urruty-sonarsource: This is also not very intentional. I think we should check what we expect

its/src/test/java/org/sonarsource/sonarqube/mcp/its/sonarcloud/tools/ListWebhooksSonarCloudIT.java
```diff
@@ -0,0 +1,40 @@
...
+class ListWebhooksSonarCloudIT extends AbstractSonarCloudStagingIT {
+
+  @Test
+  void should_call_list_webhooks_against_staging() {
+    var result = mcpClient.callTool(ListWebhooksTool.TOOL_NAME, Map.of(
+      ListWebhooksTool.PROJECT_PROPERTY, fixture.projectKey()));
+
+    assertResultEquals(result, """
+      {
+        "webhooks" : [ ]
```
damien-urruty-sonarsource, line 37: Should we have a least one webhook to check the format?

its/src/test/java/org/sonarsource/sonarqube/mcp/its/sonarcloud/tools/ListEnterprisesSonarCloudIT.java
```diff
@@ -0,0 +1,37 @@
...
+@Tag("SonarCloud")
+class ListEnterprisesSonarCloudIT extends AbstractSonarCloudStagingIT {
+
+  @Test
+  void should_call_list_enterprises_against_staging() {
+    var result = mcpClient.callTool(ListEnterprisesTool.TOOL_NAME);
+    var content = structuredContent(result);
+
+    assertThat(content).containsKey("enterprises");
+    assertThat(content.get("enterprises")).isInstanceOf(java.util.List.class);
```
damien-urruty-sonarsource: Should we check the output more precisely?

.github/workflows/build.yml
```diff
@@ -65,38 +70,67 @@ jobs:
...
+          SONARCLOUD_IT_TOKEN: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).SONARCLOUD_IT_TOKEN || '' }}
+          ARTIFACTORY_URL: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_URL || '' }}
+          ARTIFACTORY_USER: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_USER || '' }}
+          ARTIFACTORY_ACCESS_TOKEN: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).ARTIFACTORY_ACCESS_TOKEN || '' }}
+        run: |
+          if [ -z "${SONARCLOUD_IT_TOKEN}" ]; then
+            echo "::warning::SONARCLOUD_IT_TOKEN not available (e.g. fork PR); skipping SonarQube Cloud staging ITS"
+            exit 0
+          fi
+          ./gradlew :its:sonarCloudIntegrationTest --info
```
damien-urruty-sonarsource: Maybe the check for SONARCLOUD_IT_TOKEN should be part of sonarCloudIntegrationTest? So it also fails early and explicitly locally

damien-urruty-sonarsource (commented):
Looks quite good to me, I left a couple of remarks/questions

.github/workflows/sonarcloud-integration.yml
nquinquenel: Yes, I wanted to run the tests on a weekly basis (or maybe daily?) to check any sudden change on SQC - given that we may not always have activity on this repo

PR #432, opened 2026-06-04
MCP-492 analyze_code_snippet should support .tsx/.jsx

nquinquenel:
----
## Summary by Gitar

- **Language support:**
  - Updated `AnalyzeCodeSnippetTool` to support `tsx` and `jsx` as valid input languages.
- **Core analysis logic:**
  - Added `resolveAnalysisFileExtension` in `LanguageUtils` to correctly map file extensions based on language input.
  - Updated `getValidLanguageNames` to include all supported `SonarLanguage` file suffix keys.
- **Documentation:**
  - Updated `README.md` and `GEMINI.md` to reflect new language key support for JavaScript and TypeScript.

### DIFF
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -53,7 +53,7 @@ The user should provide the following environment variables:
 - Python (including IPython notebooks)
 - Ruby
 - Go
-- JavaScript, TypeScript, JSP
+- JavaScript (`js`, `jsx`), TypeScript (`ts`, `tsx`), JSP
 - PHP
 - XML
 - HTML, CSS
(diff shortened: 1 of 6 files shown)


damien-urruty-sonarsource (approved):
LGTM, although a bit complicated. I wonder if we could do things in a different way, maybe asking the agent to pass the file extension as a tool argument? Might be risky

nquinquenel:
> LGTM, although a bit complicated. I wonder if we could do things in a different way, maybe asking the agent to pass the file extension as a tool argument? Might be risky

Agree, it's quite complicated, the initial solution was slightly simpler but I got into a fight with gitar...which appeared to be valid issues. Not too sure now how to make things simpler. The previous behavior and doc was wrong and misleading. Given the low priority of this tool I'm not sure we'd like to spend more time on it 🤷🏻

PR #442, opened 2026-06-15
MCP-508 Upgrade SonarQube MCP Server to MCP Java SDK 2.0.0

nquinquenel:
----
## Summary by Gitar

- **Dependency upgrade:**
  - Upgraded `mcp-server` SDK from `1.1.3` to `2.0.0` in `libs.versions.toml` and updated lockfiles.
- **API migration:**
  - Updated `McpSchema` usage to use the new builder pattern for `Tool`, `InitializeRequest`, `CallToolRequest`, and `Implementation`.
  - Migrated `inputSchema` to use raw `Map<String, Object>` structures instead of `JsonSchema` objects.
- **Tool adjustments:**
  - Refactored `SchemaToolBuilder` to support new schema building methods and updated `SearchIssuesTool` to use `addEnumArrayProperty`.
  - Fixed property types in `SearchMyProjectsTool` and adjusted input arguments in various tests to use native types (boolean/integer) instead of strings.
- **Testing improvements:**
  - Added helper methods `assertMissingRequiredArgument` and `assertToolExecutionError` in `SonarQubeMcpTestClient` to standardize error assertions.


src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeCodeSnippetTool.java
```diff
@@ -93,7 +93,7 @@ private static McpSchema.Tool buildSchema(@Nullable String configuredProjectKey,
       .setDescription("Analyze a file or code snippet to identify code quality and security issues. " +
         "Optionally provide a code snippet to filter issues — only issues within the snippet will be reported (snippet location is auto-detected). " +
         "Always specify the language and the file scope (MAIN or TEST) for more accurate results.")
-      .addProjectKeyProperty(PROJECT_KEY_PROPERTY, "The SonarQube project key", configuredProjectKey);
+      .addOptionalProjectKeyProperty(PROJECT_KEY_PROPERTY, configuredProjectKey);
```
damien-urruty-sonarsource, line 96: Is this related to the change of SDK version? For clairty, it could have been at least a separate commit, but it's fine

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeCodeSnippetTool.java
```diff
@@ -121,9 +121,7 @@ public Result execute(Arguments arguments) {
       Thread.currentThread().interrupt();
       return handleInitializationError(e, startTime);
     }
-    var projectKey = configuredProjectKey != null
-      ? arguments.getProjectKeyWithFallback(PROJECT_KEY_PROPERTY, configuredProjectKey)
-      : arguments.getOptionalString(PROJECT_KEY_PROPERTY);
+    var projectKey = arguments.getOptionalProjectKeyWithFallback(PROJECT_KEY_PROPERTY, configuredProjectKey);
```
damien-urruty-sonarsource, line 124: It was already like this, but I find it strange that we pass `configuredProjectKey` both in the argument declaration and in the retrieval of the value. But I don't know if there is a better design here

src/main/java/org/sonarsource/sonarqube/mcp/tools/SchemaToolBuilder.java
```diff
@@ -109,7 +116,19 @@ public SchemaToolBuilder addArrayProperty(String propertyName, String itemsType,
...
-    var content = Map.of(TYPE_PROPERTY_NAME, ARRAY_TYPE, DESCRIPTION_KEY_NAME, description, ITEMS_PROPERTY_NAME, Map.of(TYPE_PROPERTY_NAME, STRING_TYPE, "enum", items));
+    var content = new HashMap<String, Object>();
+    content.put(TYPE_PROPERTY_NAME, STRING_TYPE);
+    content.put(DESCRIPTION_KEY_NAME, description);
+    content.put("enum", List.of(items));
+    properties.put(propertyName, content);
+    return this;
+  }
+
+  public SchemaToolBuilder addEnumArrayProperty(String propertyName, String[] items, String description) {
```
damien-urruty-sonarsource, line 127: What is the difference between `addEnumProperty` and `addEnumArrayProperty`? Couldn't we keep only one?

damien-urruty-sonarsource (approved):

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/AnalyzeCodeSnippetTool.java
nquinquenel: Yes indirectly, SDK 2.0 validates tool inputs against the schema before the handler runs.

`analyze_code_snippet` used `addProjectKeyProperty` which required `projectKey` when no default env, but `execute()` treated it as optional.

With 2.0 validation this mismatch would reject those calls at the MCP layer

src/main/java/org/sonarsource/sonarqube/mcp/tools/SchemaToolBuilder.java
nquinquenel: They generate different schemas that are enforced strictly by the SDK 2.0:

`addEnumProperty` -> `{ type: "string", enum: [...] }`
`addEnumArrayProperty` -> `{ type: "array", items: { type: "string", enum: [...] } }`

Before, `addEnumProperty` incorrectly emitted type: "array" for everything (bug)

damien-urruty-sonarsource: But do we need both? I don't understand the difference between the 2. I also have a vague recollection that at some point we were using the array form, but for some reason we stopped using it

nquinquenel: Yes, we have two different cases:

`"language": "java"` -> rejected if schema says `array`
`"severities": ["HIGH"]` -> rejected if schema says `string`

PR #465, opened 2026-06-24
MCP-530 Relax Origin check for authenticated and bootstrap requests

nquinquenel:
----
## Summary by Gitar

- **Security updates:**
  - Relaxed `Origin` header enforcement in `McpSecurityFilter` for authenticated requests, OAuth bootstrap paths (`/mcp`), and `/.well-known/*` endpoints.
  - Remote host bindings no longer enforce strict `Origin` checks.
- **Documentation:**
  - Updated `http-authentication-architecture.md` to reflect new origin validation logic and clarify browser-layer backstop role.
- **Testing:**
  - Added integration tests to verify authenticated request bypass and enforcement on non-MCP paths.
  - Updated unit tests to cover new bypass conditions for bootstrap paths and tokens.

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/transport/McpSecurityFilter.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/transport/McpSecurityFilter.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.List;
 import java.util.Set;
+import javax.annotation.Nullable;
 import org.sonarsource.sonarqube.mcp.authentication.AuthenticationFilter;
 import org.sonarsource.sonarqube.mcp.log.McpLogger;
 
@@ -98,7 +99,7 @@ public class McpSecurityFilter implements Filter {
     var origin = httpRequest.getHeader("Origin");
     boolean isOptionsRequest = "OPTIONS".equals(httpRequest.getMethod());
 
-    if (origin != null && !origin.isBlank() && shouldEnforceOrigin(httpRequest) && !isOriginAllowed(origin)) {
+    if (!isOriginAllowed(httpRequest, origin)) {
       LOG.warn("Rejected request from disallowed origin: " + origin);
       httpResponse.setStatus(HttpServletResponse.SC_FORBIDDEN);
       httpResponse.setContentType("application/json");
@@ -106,7 +107,7 @@ public class McpSecurityFilter implements Filter {
       return;
     }
 
-    if (origin != null && isOriginAllowed(origin)) {
+    if (origin != null && isOriginInAllowlist(origin)) {
       httpResponse.setHeader("Access-Control-Allow-Origin", origin);
     } else if (isOptionsRequest) {
       httpResponse.setHeader("Access-Control-Allow-Origin", "*");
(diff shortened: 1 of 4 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/transport/McpSecurityFilter.java
```diff
@@ -91,7 +98,7 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain filte
     var origin = httpRequest.getHeader("Origin");
     boolean isOptionsRequest = "OPTIONS".equals(httpRequest.getMethod());
 
-    if (origin != null && !isOriginAllowed(origin)) {
+    if (origin != null && !origin.isBlank() && shouldEnforceOrigin(httpRequest) && !isOriginAllowed(origin)) {
```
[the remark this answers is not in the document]
nquinquenel: Intentional. Blank Origin is treated like absent, browsers don't send Origin: "" on cross-site requests, and non-browser clients aren't in the DNS rebinding threat model. Rejecting blank was stricter than needed and inconsistent with the existing no-Origin pass-through. Downstream auth still applies; no tool execution without a token.

src/main/java/org/sonarsource/sonarqube/mcp/transport/McpSecurityFilter.java
```diff
@@ -135,6 +142,37 @@ private static String escapeJson(String value) {
...
+      return false;
+    }
+    return !isOAuthBootstrapRequest(request);
+  }
+
+  private boolean isLocalBinding() {
+    return "127.0.0.1".equals(hostBinding) || "localhost".equals(hostBinding) || "0.0.0.0".equals(hostBinding);
+  }
+
+  private static boolean hasAuthToken(HttpServletRequest request) {
```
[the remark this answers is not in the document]
nquinquenel: The bootstrap exemption is intentional and relies on AuthenticationFilter rejecting tokenless /mcp in TOKEN mode. Added a test to lock that in: disallowed Origin + no token → 401, not 403.

src/main/java/org/sonarsource/sonarqube/mcp/transport/McpSecurityFilter.java
```diff
@@ -135,6 +142,38 @@ private static String escapeJson(String value) {
...
+      return false;
+    }
+    if (hasAuthToken(request)) {
+      return false;
+    }
+    return !isOAuthBootstrapRequest(request);
+  }
+
+  private boolean isLocalBinding() {
+    return "127.0.0.1".equals(hostBinding) || "localhost".equals(hostBinding) || "0.0.0.0".equals(hostBinding);
```
damien-urruty-sonarsource, line 184: IP V6? There are also plenty of other valid loopback addresses. I think the standard way in Java is to use `InetAddress.isLoopbackAddress`

src/main/java/org/sonarsource/sonarqube/mcp/transport/McpSecurityFilter.java
```diff
@@ -135,6 +142,38 @@ private static String escapeJson(String value) {
     return value.replace("\\", "\\\\").replace("\"", "\\\"");
   }
 
+  /**
+   * Whether to reject a disallowed {@code Origin} header.
+   * Skipped for authenticated requests, OAuth bootstrap paths, and remote deployments.
+   */
+  private boolean shouldEnforceOrigin(HttpServletRequest request) {
```
damien-urruty-sonarsource: I find the name a bit blurry, not easy to understand when read in the condition above. Couldn't we keep a single `isOriginAllowed`, that would check all conditions? Maybe even including null and empty checks. There are also some negations, which makes it harder to reason about in this form

src/main/java/org/sonarsource/sonarqube/mcp/transport/McpSecurityFilter.java
nquinquenel: Given the feedback from Gitar, `isLoopbackAddress` seems like an issue in regards of performance and security risks 👇🏻

damien-urruty-sonarsource (approved):
LGTM, but maybe a review from IST or AppSec would help

PR #484, opened 2026-07-06
MCP-535 Add SBOM and max-mode provenance

nquinquenel:
----
## Summary by Gitar

- **CI/CD configuration:**
  - Enabled SBOM generation and set provenance to `mode=max` in `docker-publish.yml` for container builds.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -77,6 +77,8 @@ jobs:
           context: .
           platforms: linux/amd64,linux/arm64
           push: true
+          sbom: true
+          provenance: mode=max
           build-args: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -178,6 +180,8 @@ jobs:
           context: .
           platforms: linux/amd64,linux/arm64
           push: true
+          sbom: true
+          provenance: mode=max
```
damien-urruty-sonarsource, line 184: Is it normal that we are building the image twice? It looks strange to me that we run tests against an image that is not the one that will be published

damien-urruty-sonarsource (approved):

PR #491, opened 2026-07-08
MCP-550 Auto-detect SonarQube Cloud Organization (Single Org)

nquinquenel:
----
## Summary by Gitar

- **Organization Auto-detection:**
  - Added `autoDetectOrganizationIfNeeded` to resolve the organization via the token in stdio mode when `SONARQUBE_ORG` is unset.
  - Added validation to handle cases with zero or multiple organizations, providing actionable error messages.
- **API Changes:**
  - Updated `OrganizationsApi.listOrganizations` to return a `List<Organization>` and propagate HTTP errors.
- **Configuration Updates:**
  - Removed strict validation requiring `SONARQUBE_ORG` in stdio Cloud mode to support automatic detection.
- **Testing:**
  - Added `AutoDetectOrganizationTest` to verify auto-selection logic and error handling.
  - Updated `OrganizationsApiTest` to cover the new listing functionality.

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
@@ -54,6 +54,7 @@ import org.sonarsource.sonarqube.mcp.serverapi.ServerApiHelper;
 import org.sonarsource.sonarqube.mcp.serverapi.ServerApiProvider;
 import org.sonarsource.sonarqube.mcp.serverapi.features.Feature;
 import org.sonarsource.sonarqube.mcp.serverapi.organizations.OrganizationsApi;
+import org.sonarsource.sonarqube.mcp.serverapi.organizations.ResolvedOrganization;
 import org.sonarsource.sonarqube.mcp.slcore.BackendService;
 import org.sonarsource.sonarqube.mcp.tools.Tool;
 import org.sonarsource.sonarqube.mcp.tools.ToolCategory;
@@ -176,14 +177,11 @@ public class SonarQubeMcpServer implements ServerApiProvider {
   @Nullable
   private ServerApi serverApi;
   /**
-   * Effective SonarQube Cloud organization used by the stdio startup path.
-   * Initialized from {@code SONARQUBE_ORG}; when unset in stdio Cloud mode it may be auto-detected from the token
-   * (see {@link #autoDetectOrganizationIfNeeded()}). The HTTP path resolves the organization per request instead.
+   * Effective SonarQube Cloud organization for the stdio session (key and optional cached UUID v4).
+   * Set by {@link #resolveOrganizationAtStartup()}. HTTP mode resolves organization per request instead.
    */
   @Nullable
-  private String resolvedOrganization;
-  @Nullable
-  private String resolvedOrganizationUuidV4;
+  private ResolvedOrganization resolvedOrganization;
   private SonarQubeVersionChecker sonarQubeVersionChecker;
   @Nullable
   private McpStatelessSyncServer statelessSyncServer;
@@ -283,13 +281,8 @@ public class SonarQubeMcpServer implements ServerApiProvider {
       ? new ToolExecutor(backendService, analyticsService, null, this, mcpConfiguration.getMcpServerId())
       : new ToolExecutor(backendService, analyticsService, connectionContext, null, mcpConfiguration.getMcpServerId());
 
-    this.resolvedOrganization = mcpConfiguration.getSonarqubeOrg();
-
-    // Create ServerApi for startup probing (version check, SCA availability, plugin sync).
-    // In HTTP mode this is optional — only created when a startup token is configured.
-    // Per-request tool calls in HTTP mode always use a fresh ServerApi from the request headers.
     this.serverApi = initializeServerApi(mcpConfiguration);
-    autoDetectOrganizationIfNeeded();
+    resolveOrganizationAtStartup();
     this.sonarQubeVersionChecker = new SonarQubeVersionChecker(serverApi);
     loadBackendIndependentTools(serverApi);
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
```diff
@@ -484,6 +497,14 @@ private static boolean isCagEnabledForOrg(@Nullable ServerApi api, @Nullable Str
     return entitlement.allowed();
   }
 
+  @Nullable
+  private String organizationUuidV4(ServerApi api, String orgKey) {
+    if (orgKey.equals(resolvedOrganization) && resolvedOrganizationUuidV4 != null) {
```
georgii-borovinskikh-sonarsource: it looks like this method is always called with the resolvedOrganization - unless I'm missing something

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
```diff
@@ -695,9 +716,46 @@ private ServerApi initializeServerApi(McpServerLaunchConfiguration mcpConfigurat
...
   }
-  
+
+  /**
+   * In stdio mode connecting to SonarQube Cloud without an explicit {@code SONARQUBE_ORG}, resolves the
+   * organization from the authenticated token. When the token belongs to exactly one organization it is
+   * adopted automatically and the startup {@link ServerApi} is rebuilt with it. Zero or multiple organizations
+   * fail startup with an actionable message (multi-org selection tooling is tracked separately in MCP-551).
+   */
+  private void autoDetectOrganizationIfNeeded() {
```
georgii-borovinskikh-sonarsource: This file is already giant, personally I would move it somewhere else

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
```diff
@@ -273,10 +283,13 @@ private void initializeBasicServicesAndTools() {
       ? new ToolExecutor(backendService, analyticsService, null, this, mcpConfiguration.getMcpServerId())
       : new ToolExecutor(backendService, analyticsService, connectionContext, null, mcpConfiguration.getMcpServerId());
 
+    this.resolvedOrganization = mcpConfiguration.getSonarqubeOrg();
```
georgii-borovinskikh-sonarsource: this assignment should happen in the autodetect method (or be the return value of it). Currently it's hard to track when this field changes it's value

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/RunAdvancedCodeAnalysisTool.java
```diff
@@ -75,13 +75,13 @@ private static McpSchema.Tool buildSchema(@Nullable String configuredProjectKey)
       .build();
   }
 
-  public static boolean isA3sEnabled(ServerApi api, String orgKey) {
-    var orgUuidV4 = api.organizationsApi().getOrganizationUuidV4(orgKey);
-    if (orgUuidV4 == null) {
+  public static boolean isA3sEnabled(ServerApi api, String orgKey, @Nullable String orgUuidV4) {
+    var uuid = orgUuidV4 != null ? orgUuidV4 : api.organizationsApi().getOrganizationUuidV4(orgKey);
```
georgii-borovinskikh-sonarsource: the checks to get Uuid all over the place look a bit hacky. Is there a way to do it once and cache? Maybe create an object to represent the org key + uuid

src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
```diff
@@ -422,17 +422,20 @@ private static void validateStdioConfiguration(boolean isHttpEnabled, @Nullable
    *   1. SONARQUBE_IS_CLOUD=true → explicit override
    *   2. SONARQUBE_ORG set → org implies SQC
    *   3. SONARQUBE_URL is a known SQC hostname (sonarcloud.io, sonarqube.us, …)
-   *   4. No SONARQUBE_URL → default to SQC (HTTP mode only after stdio validation)
+   *   4. No SONARQUBE_URL → default to SQC (HTTP mode only)
```
georgii-borovinskikh-sonarsource: can you explain this part?

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
nquinquenel: Yes, we have a ticket dedicated for this, for now I'd rather not include this change in scope of this PR

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
nquinquenel: You're right, all startup entitlement checks pass `resolvedOrganization`, so the `orgKey.equals` was redundant. I replaced this by a `ResolvedOrganization` record, I think this simplifies things

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
nquinquenel: Fair point, but this is required here because the org key must be in `EndpointParams` before the first `ServerApi` is created. It's a bit complicated but long story short, I'm not sure we can do better unless huge refactoring.

src/main/java/org/sonarsource/sonarqube/mcp/tools/analysis/RunAdvancedCodeAnalysisTool.java
nquinquenel: Agree, I modified this to leverage a new `ResolvedOrganization` record

src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
nquinquenel: The old wording was misleading.

Rule 4 means: if there's no `SONARQUBE_URL`, we only treat the instance as Cloud when HTTP transport is enabled (HTTP defaults the URL to sonarcloud.io). In stdio, we don't default to Cloud without a URL. `validateStdioConfiguration` still requires `SONARQUBE_URL` or `SONARQUBE_ORG`, so a user targeting self-hosted Server without a URL fails fast instead of sending their token to sonarcloud.io.

georgii-borovinskikh-sonarsource (approved):

PR #494, opened 2026-07-09
MCP-523 Allow MCP to configure client certificate

Krosovok:
----
## Summary by Gitar

- **Security/TLS Configuration:**
  - Updated `HttpClientProvider` to support client certificate authentication using system properties `javax.net.ssl.keyStore` and `trustStore`.
  - Modified `docker-entrypoint.sh` to propagate `JAVA_OPTS` to the running JVM instance.
- **Testing:**
  - Added `HttpClientProviderClientCertificateTests` to verify mTLS handshake scenarios and invalid keystore handling.
- **Documentation:**
  - Added comprehensive mTLS configuration guide to `README.md` for both Docker and standalone JAR deployments.
  - Updated `src/test/resources/ssl/README.md` to clarify the dual usage of test certificates in mTLS scenarios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -850,6 +850,8 @@ When using custom certificates, you can modify your MCP configuration to mount t
 }
 ```
 
+> **Note:** Running the server [from a JAR](#build) instead of the container? The volume mount above installs certificates into the container's OS trust store, which the server also reads. If you cannot use the OS trust store — notably on Windows, where it is not consulted — point the JVM at a Java truststore holding the CA certificate: `-Djavax.net.ssl.trustStore=/path/to/truststore.p12 -Djavax.net.ssl.trustStoreType=PKCS12 -Djavax.net.ssl.trustStorePassword=<passphrase>`. It is added on top of the default trusted certificates.
+
 </details>
 
 ### Proxy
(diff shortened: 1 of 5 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/http/HttpClientProvider.java
```diff
@@ -44,7 +44,8 @@ public class HttpClientProvider {
   public HttpClientProvider(String userAgent) {
     this.userAgent = userAgent;
     var sslFactoryBuilder = SSLFactory.builder()
-      .withDefaultTrustMaterial();
+      .withDefaultTrustMaterial()
+      .withSystemPropertyDerivedIdentityMaterial();
```
nquinquenel: I think we should add a try/catch around to avoid breakage for all users, and also a check on the property:
```
if (System.getProperty("javax.net.ssl.keyStore") != null) {
  try {
    sslFactoryBuilder.withSystemPropertyDerivedIdentityMaterial();
  } catch (Exception e) {
    LOG.warn("Could not load client certificate identity material: " + e.getMessage());
  }
}
```

README.md
```diff
@@ -895,6 +895,55 @@ SOCKS5 proxies are supported.
...
+{
+  "sonarqube": {
+    "command": "docker",
+    "args": [
+      "run", "--init", "--pull=always", "-i", "--rm",
+      "-v", "/path/to/client.p12:/etc/ssl/mcp/client.p12:ro",
+      "-e", "JAVA_OPTS",
+      "-e", "SONARQUBE_TOKEN",
+      "-e", "SONARQUBE_URL",
+      "mcp/sonarqube"
```
nquinquenel:
```suggestion
      "sonarsource/sonarqube-mcp"
```

README.md
```diff
@@ -895,6 +895,55 @@ SOCKS5 proxies are supported.
...
+  -v /path/to/client.p12:/etc/ssl/mcp/client.p12:ro \
+  -e JAVA_OPTS="-Djavax.net.ssl.keyStore=/etc/ssl/mcp/client.p12 -Djavax.net.ssl.keyStoreType=PKCS12 -Djavax.net.ssl.keyStorePassword=<passphrase>" \
+  -e SONARQUBE_TOKEN="<token>" \
+  -e SONARQUBE_URL="<url>" \
+  mcp/sonarqube
+```
+
+Omit `-Djavax.net.ssl.keyStorePassword` if the keystore has no passphrase.
+
+#### MCP Configuration with a Client Certificate
```
nquinquenel, line 928: Should we provide a configuration example for the Jar as well (here we only mention via Docker)

nquinquenel (commented):
Is there any test worth adding?

README.md
```diff
@@ -895,6 +895,55 @@ SOCKS5 proxies are supported.
...
+
+```json
+{
+  "sonarqube": {
+    "command": "docker",
```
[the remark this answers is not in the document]
Krosovok: I don't think it's worth documenting. The goal is to show how the system behaves / where does it take values from. The main consumer is going to be our own SonarQube CLI that would provide the values dynamically at runtime, without the file.

src/main/java/org/sonarsource/sonarqube/mcp/http/HttpClientProvider.java
Krosovok: Yes, I am on it. Just jumping between the two.

README.md
```diff
@@ -895,6 +895,90 @@ SOCKS5 proxies are supported.
...
+
+Mount your `.p12` or `.pfx` file into the container and set the `JAVA_OPTS` environment variable with the keystore properties:
+
+```bash
+docker run --init --pull=always -i --rm \
+  -v /path/to/client.p12:/etc/ssl/mcp/client.p12:ro \
+  -e JAVA_OPTS="-Djavax.net.ssl.keyStore=/etc/ssl/mcp/client.p12 -Djavax.net.ssl.keyStoreType=PKCS12 -Djavax.net.ssl.keyStorePassword=<passphrase>" \
+  -e SONARQUBE_TOKEN="<token>" \
+  -e SONARQUBE_URL="<url>" \
+  mcp/sonarqube
```
nquinquenel:
```suggestion
  sonarsource/sonarqube-mcp
```

src/test/java/org/sonarsource/sonarqube/mcp/http/HttpClientProviderTests.java
```diff
@@ -35,19 +40,79 @@
...
+    var underTest = new HttpClientProvider(USER_AGENT);
+
+    try (var ignored = underTest.getHttpClient("token").getAsync(sonarqubeMock.url("/test")).join()) {
+      // nothing
+    }
+    sonarqubeMock.verify(getRequestedFor(urlEqualTo("/test")));
+  }
+
+  @Test
+  void it_should_create_client_when_keystore_property_points_to_valid_keystore() throws Exception {
```
nquinquenel: In this test, WireMock isn't configured to require a client certificate, so I'm not sure this test is actually meaningful

nquinquenel (approved):

src/test/java/org/sonarsource/sonarqube/mcp/http/HttpClientProviderTests.java
Krosovok: I ended up creating a separate test with mTLS setup. I had to add CA cert params as well so that I can use them in test. Please take another look!

src/main/java/org/sonarsource/sonarqube/mcp/http/HttpClientProvider.java
```diff
@@ -45,13 +45,19 @@ public HttpClientProvider(String userAgent) {
     this.userAgent = userAgent;
     var sslFactoryBuilder = SSLFactory.builder()
       .withDefaultTrustMaterial();
+    if (isClientCertificateConfigured()) {
+      sslFactoryBuilder.withSystemPropertyDerivedIdentityMaterial();
+    }
```
nquinquenel, line 51: A failure at this step would crash the server, we usually try not to fail hard on such cases. I think it's worth catching with a warning log.

src/main/java/org/sonarsource/sonarqube/mcp/http/HttpClientProvider.java
```diff
@@ -45,13 +45,19 @@ public HttpClientProvider(String userAgent) {
...
     if (!SystemUtils.IS_OS_WINDOWS) {
       try {
         sslFactoryBuilder.withSystemTrustMaterial();
       } catch (Exception e) {
         LOG.warn("Could not load system trust material, falling back to JDK defaults: " + e.getMessage());
       }
     }
+    if (isCustomTrustStoreConfigured()) {
+      sslFactoryBuilder.withSystemPropertyDerivedTrustMaterial();
+    }
```
nquinquenel, line 59: Same here

README.md
```diff
@@ -895,6 +897,90 @@ SOCKS5 proxies are supported.
...
+<summary>Configuration</summary>
+
+#### Using a PKCS12 keystore
+
+Mount your `.p12` or `.pfx` file into the container and set the `JAVA_OPTS` environment variable with the keystore properties:
+
+```bash
+docker run --init --pull=always -i --rm \
+  -v /path/to/client.p12:/etc/ssl/mcp/client.p12:ro \
+  -e JAVA_OPTS="-Djavax.net.ssl.keyStore=/etc/ssl/mcp/client.p12 -Djavax.net.ssl.keyStoreType=PKCS12 -Djavax.net.ssl.keyStorePassword=<passphrase>" \
```
nquinquenel, line 914: `keyStorePassword=<passphrase>` would leak into the process list or in docker inspect. It's a limitation from the JVM system-property approach but I'd still add a note somewhere, as the passphrase would be disclosed to anyone

PR #510, opened 2026-07-31
feat: make background initialization timeout configurable

### LINKED ISSUE #508: Make background initialization timeout an optional configurable value
### Problem / use case

Currently, in environments with significant latency, it is not possible to manually run the MCP container because it shuts down before it is able to download all of the plugins from the server. Some of the downloads complete successfully but due to the downloads taking longer than expected, the shutdown sequence is started prior to the other plugins being given enough time to download. See logs below:

```
DEBUG SonarQube MCP Server - === Debug Level Configuration Details ===
DEBUG SonarQube MCP Server - SSL/TLS - OS: Linux
DEBUG SonarQube MCP Server - SSL/TLS configured - protocol: TLS, trusted certificates: 263
DEBUG SonarQube MCP Server - Proxy selector: sun.net.spi.DefaultProxySelector
DEBUG SonarQube MCP Server - No proxy system properties configured
DEBUG SonarQube MCP Server - HTTP client user agent: SonarQube MCP Server 1.22.0.3040
DEBUG SonarQube MCP Server - Enabled toolsets: [DUPLICATIONS, PROJECTS, MEASURES, ISSUES, DEPENDENCY_RISKS, CAG, COVERAGE, RULES, AGENTIC_READINESS, SECURITY_HOTSPOTS, QUALITY_GATES, ANALYSIS]
DEBUG SonarQube MCP Server - Advanced analysis: false
DEBUG SonarQube MCP Server - Telemetry enabled: true
DEBUG SonarQube MCP Server - App version: 1.22.0.3040
DEBUG SonarQube MCP Server - Storage path: /app/storage
DEBUG SonarQube MCP Server - Log file: /app/storage/logs/mcp.log
DEBUG SonarQube MCP Server - IDE port: not set
DEBUG SonarQube MCP Server - ================================
INFO SonarQube MCP Server - Downloading analyzers in background...
INFO SonarQube MCP Server - Successfully downloaded plugin 'go' to /app/storage/plugins/sonar-go-enterprise-plugin-1.37.0.6200.jar
INFO SonarQube MCP Server - Successfully downloaded plugin 'web' to /app/storage/plugins/sonar-html-plugin-3.26.0.7600.jar
INFO SonarQube MCP Server - Successfully downloaded plugin 'iac' to /app/storage/plugins/sonar-iac-enterprise-plugin-2.10.0.20406.jar
INFO SonarQube MCP Server - Successfully downloaded plugin 'javasymbolicexecution' to /app/storage/plugins/sonar-java-symbolic-execution-plugin-8.21.0.1991.jar
WARN SonarQube MCP Server - Background initialization did not complete within 30 seconds, proceeding with shutdown
ERROR SonarQube MCP Server - Failed to download plugin 'java'
```

### Proposed solution

It would be helpful to expand the timeout window to an appropriate amount of time for the environment and using the 30 seconds as a default if none is provided. I propose making the timeout value an optional value set in an environment variable such as `MCP_INIT_TIMEOUT` or something more appropriate. If this variable is not defined, the default could be 30 seconds

### Alternatives considered

I would prefer not to go down this route but disabling the timeout all together and letting the plugin download failures cause the shutdown instead

Slacknsss:
Closes #508

## Problem

On a high-latency connection the container shuts down before it has finished downloading its analyzers. `awaitBackgroundInitialization()` waits a hardcoded 30 seconds, then cancels the future, and the remaining plugin downloads fail:

```
WARN  Background initialization did not complete within 30 seconds, proceeding with shutdown
ERROR Failed to download plugin 'java'
```

## Change

Adds `SONARQUBE_INIT_TIMEOUT_SECONDS`, defaulting to 30, so behaviour is unchanged when it is not set.

Two deliberate choices, both open to discussion:

**Naming.** @elegier suggested `MCP_INIT_TIMEOUT`. I used `SONARQUBE_INIT_TIMEOUT_SECONDS` instead, to match the `SONARQUBE_*` convention every other variable in `McpServerLaunchConfiguration` follows, and to put the unit in the name rather than leave it implicit. Happy to rename if you prefer the original.

**Not the alternative.** The issue also floated removing the timeout entirely and letting download failures drive shutdown. I did not do that — an unbounded wait on a shutdown path can hang a container indefinitely, which is a worse failure than a slow start.

Parsing mirrors the existing `parseHttpPortValue`: blank falls back to the default, non-numeric and non-positive values fail fast with an explicit message.

The warning itself said "within 30 seconds" regardless of the configured value, so it now reports the real timeout and names the variable to raise.

`README.md` documents the variable in a short "Startup and shutdown" table, placed before "Transport Modes" since it is not transport-specific.

## Tests

Five tests added to `McpServerLaunchConfigurationTest`: default, explicit value, blank falls back to default, non-numeric rejected, non-positive rejected.

## ⚠️ I could not run your build, and I think that is worth reporting on its own

`./gradlew build` fails for anyone outside SonarSource. `build.gradle.kts` falls back to `mavenCentral()` when Artifactory credentials are absent, but the pinned `sonarlint-core` artifacts are not there:

- `org.sonarsource.sonarlint.core:sonarlint-core:11.8.0.86062` → **404** on Maven Central (it only publishes up to `10.24.0.81415`)
- the same coordinates on `repox.jfrog.io` → **401**

So an external contributor cannot compile or test this repository at all. That is a real barrier to the outside contributions the `good first issue` and `help wanted` labels invite.

What I did instead: extracted `parseInitTimeoutSecondsValue` verbatim into a standalone file and ran it under Java 21 (the toolchain version this project targets) across all seven cases — default from null, default from blank, explicit value, lower bound of 1, non-numeric, zero, negative. All seven behave as intended. That validates the parsing logic, **not** the integration, so please let CI be the judge on the wiring.

I would rather flag this than quietly present the change as fully verified.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -647,6 +647,12 @@ To enable full functionality, the following environment variables must be set be
 
 > 💡 **Configuration Tip (stdio mode)**: The presence of `SONARQUBE_ORG` determines whether you're connecting to SonarQube Cloud or Server. If `SONARQUBE_ORG` is set, SonarQube Cloud is used; otherwise, SonarQube Server is used.
 
+#### Startup and shutdown
+
+| Environment variable             | Description                                                                                                                                                    | Default |
+|----------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
+| `SONARQUBE_INIT_TIMEOUT_SECONDS` | How long shutdown waits for background initialization (analyzer downloads) to finish. Raise it on slow or high-latency connections where analyzers time out. | `30`    |
+
 ### Transport Modes
 
 The [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports) defines two transport mechanisms: **Stdio** and **Streamable HTTP**. The SonarQube MCP Server supports both:
--- a/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
@@ -811,11 +811,13 @@ public class SonarQubeMcpServer implements ServerApiProvider {
     if (initializationFuture.isDone()) {
       return;
     }
+    var timeoutSeconds = mcpConfiguration.getInitTimeoutSeconds();
     LOG.info("Waiting for background initialization to complete before shutdown...");
     try {
-      initializationFuture.get(30, TimeUnit.SECONDS);
+      initializationFuture.get(timeoutSeconds, TimeUnit.SECONDS);
     } catch (TimeoutException | ExecutionException e) {
-      LOG.warn("Background initialization did not complete within 30 seconds, proceeding with shutdown");
+      LOG.warn("Background initialization did not complete within " + timeoutSeconds
+        + " seconds, proceeding with shutdown. Set SONARQUBE_INIT_TIMEOUT_SECONDS to allow more time.");
       initializationFuture.cancel(true);
     } catch (Exception e) {
       LOG.error("Background initialization failed or was interrupted", e);
--- a/src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
@@ -95,6 +95,10 @@ public class McpServerLaunchConfiguration {
 
   private static final String SONARQUBE_MCP_IN_CONTAINER = "SONARQUBE_MCP_IN_CONTAINER";
 
+  // Shutdown configuration
+  private static final String SONARQUBE_INIT_TIMEOUT_SECONDS = "SONARQUBE_INIT_TIMEOUT_SECONDS";
+  private static final int DEFAULT_INIT_TIMEOUT_SECONDS = 30;
+
   private final Path storagePath;
   private final String hostMachineAddress;
   private final String sonarqubeUrl;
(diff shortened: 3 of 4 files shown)


Slacknsss:
Closing this one. `AGENTS.md` says external pull requests are not accepted, and I should have read it before opening the PR rather than after. My mistake, and sorry for the review noise.

Two things worth keeping if anyone picks up #508 internally.

**The fix itself.** `awaitBackgroundInitialization()` waits a hardcoded 30 seconds, then cancels the future, so on a high-latency connection the analyzer downloads fail and the server starts without them. An environment variable defaulting to 30 leaves current behaviour untouched. On naming, @elegier suggested `MCP_INIT_TIMEOUT`; I would argue for `SONARQUBE_INIT_TIMEOUT_SECONDS`, to match the convention every other variable in `McpServerLaunchConfiguration` already follows and to put the unit in the name, but that is a house-style call. The other option floated on the issue — dropping the timeout entirely — looks worse to me: an unbounded wait on a shutdown path can hang a container indefinitely, which is a harder failure to diagnose than a slow start.

**A separate finding, independent of this PR.** `./gradlew build` cannot succeed outside SonarSource. `build.gradle.kts` falls back to `mavenCentral()` when Artifactory credentials are absent, but the pinned `sonarlint-core` artifacts are not published there — 404 on Maven Central, 401 on repox. So nobody outside the organisation can run the test suite, including on a fork. That may well be intentional given the contribution policy, but it is not stated anywhere and it is worth knowing that forks are build-broken by default.

PR #525, opened 2026-08-10
MCP-571 MCP-570 Introduce the unified `vortex` toolset and split `ide` toolset

Krosovok:
----
## Summary by Gitar

- **Toolset Updates:**
    - Introduced unified `vortex` toolset combining CAG and A3S with shared entitlement checks in `OrgFeatureEntitlements.java`
    - Split IDE bridge analysis tools into a dedicated `ide` toolset and added it to default enabled categories


src/main/java/org/sonarsource/sonarqube/mcp/serverapi/OrgFeatureEntitlements.java
```diff
@@ -0,0 +1,95 @@
...
+      return false;
+    }
+    var orgKey = org.key();
+    var orgUuidV4 = org.uuidV4() != null ? org.uuidV4() : api.organizationsApi().getOrganizationUuidV4(orgKey);
+    if (orgUuidV4 == null) {
+      LOG.debug("Entitlement check: could not resolve UUID for org '" + orgKey + "' - skipping");
+      return false;
+    }
+    return check.test(orgUuidV4, orgKey);
+  }
```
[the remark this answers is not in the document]
Krosovok: Yes and I deliberately decided not to bring it back for now:
1. The impact is indeed minimal - saving 1 call iff both tools are enabled (it was at least 2 when cag and a3s were separate)
2. Together with existing `ResolvedOrganization` fallback it will create quite multi-step route to get one value.

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/agenticreadiness/WasFeatureFlagsApi.java
```diff
@@ -47,6 +49,9 @@ public boolean isAgenticReadinessAssessmentEnabled(String organizationId) {
     try (var response = helper.getApiSubdomain(path)) {
       var flags = GSON.fromJson(response.bodyAsString(), FeatureFlagsResponse.class);
       return flags != null && flags.isAgenticReadinessAssessmentEnabled();
+    } catch (Exception e) {
```
georgii-borovinskikh-sonarsource, line 52: I think this change doesn't belong to this PR - it's about SARA, not SQAA

Krosovok: It had the same check-for-nulls, check-for-uuid logic as both CAG and A3S so I moved that common part from `SonarQubeMcpServer` to `OrgFeatureEntitlements`. Had to align level where the exception is handled so that I wouldn't need additional logic to handle exceptions differently.

georgii-borovinskikh-sonarsource (approved):
1) We might need to separately check for entitlement if the server doesn't return that info in the object
2) We need to warn the user about deprecation of CAG and Analysis toolsets

Krosovok:
> 1. We might need to separately check for entitlement if the server doesn't return that info in the object
> 2. We need to warn the user about deprecation of CAG and Analysis toolsets

1. If there is a need, we can do it in the scope of a separate tiket.
2. Added to readme, logs and instructions. Not sure where else it can be stated.

PR #529, opened 2026-08-13
MCP-575 Add matching toolsets to invocation telemetry

tomshafir-sonarsource:
----
## Summary by Gitar

- **Telemetry enhancement:**
    - Added matching toolsets to `McpToolInvokedEvent` analytics telemetry and bumped event version to `2`
    - Updated `ToolExecutor` and `SonarQubeMcpServer` to resolve and pass enabled toolsets for tool invocations

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
@@ -282,8 +282,8 @@ public class SonarQubeMcpServer implements ServerApiProvider {
     // In stdio mode: pass the shared pre-resolved ConnectionContext
     // In HTTP mode: pass a supplier that captures the request-scoped ServerApi synchronously to the async analytics task
     this.toolExecutor = mcpConfiguration.isHttpEnabled()
-      ? new ToolExecutor(backendService, analyticsService, null, this, mcpConfiguration.getMcpServerId())
-      : new ToolExecutor(backendService, analyticsService, connectionContext, null, mcpConfiguration.getMcpServerId());
+      ? new ToolExecutor(backendService, analyticsService, null, this, mcpConfiguration.getMcpServerId(), this::resolveEnabledToolsets)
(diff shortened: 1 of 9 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
```diff
@@ -581,6 +581,24 @@ private McpServerFeatures.SyncToolSpecification toStdioSpec(Tool tool) {
       .build();
   }
 
+  @VisibleForTesting
+  Set<ToolCategory> resolveEnabledToolsets() {
+    var enabledToolsets = EnumSet.noneOf(ToolCategory.class);
+    ToolCategory.all().stream()
+      .filter(mcpConfiguration::isToolCategoryEnabled)
+      .forEach(enabledToolsets::add);
```
Krosovok: I think that when the goal to create a collection it is more intentional to use `collect`/`toSet`/`toList`/etc and keep `forEach` for cases where there is more complex final processing to be done on elements.
```suggestion
       var enabledToolsets = ToolCategory.all().stream()
      .filter(mcpConfiguration::isToolCategoryEnabled)
      .collect(Collectors.toCollection(() -> EnumSet.noneOf(ToolCategory.class)));
```

src/main/java/org/sonarsource/sonarqube/mcp/SonarQubeMcpServer.java
```diff
@@ -581,6 +581,24 @@ private McpServerFeatures.SyncToolSpecification toStdioSpec(Tool tool) {
...
+
+    var transportContext = currentTransportContext.get();
+    if (transportContext != null) {
+      var requestToolsets = transportContext.get(HttpServerTransportProvider.CONTEXT_TOOLSETS_KEY);
+      if (requestToolsets instanceof Set<?> toolsets) {
+        enabledToolsets.removeIf(toolset -> !toolsets.contains(toolset));
```
Krosovok: Similarly, more intentional for set intersection.
```suggestion
        enabledToolsets.retainAll(toolsets);
```

Krosovok (approved):
LGTM, left some nitpick comments

PR #530, opened 2026-08-13
MCP-572 Deprecate `analyze_code_snippet` tool

Krosovok:
The tool is already loaded only if both vortex and ide tools are not possible, just adding notes about deprecation into response and documentation.


Krosovok:
@gitar-bot it `AnalyzeCodeSnippetTool` already loaded only when Vortex is disabled.

tomshafir-sonarsource (approved):

PR #538, opened 2026-08-24
MCP-608 Answer MCP initialize with an error on stdio startup failure

nquinquenel:
image

## Summary
- On stdio startup failure (missing token, URL, storage path, …), wait for the client’s MCP `initialize` request and reply with a JSON-RPC error instead of exiting silently.
- Clients can show the real cause instead of `Connection closed`.
- HTTP/HTTPS launches skip this stdin wait; they only log the failure.
- Uses existing MCP JSON-RPC types (`McpSchema` / `McpJsonMappers`), not a second protocol stack.

## Test plan
- [x] Stdio with missing `SONARQUBE_TOKEN`: client receives an `initialize` error containing the token message
- [x] HTTP/HTTPS startup failure: process does not block on stdin

----
## Summary by Gitar

- **Error reporting:**
    - Added `StdioInitializeErrorReporter` to catch startup failures and reply to `initialize` with JSON-RPC errors instead of closing connection
    - Created `JsonRpcErrorResponses` utility for generating JSON-RPC 2.0 error payloads with code `-32000`
    - Updated `SonarQubeMcpServer` main method to catch runtime exceptions and report initialization errors via stdio

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
@@ -150,9 +150,8 @@ public class McpServerLaunchConfiguration {
     this.storagePath = Paths.get(storagePathString);
     this.hostMachineAddress = resolveHostMachineAddress();
 
-    var transportMode = requireNonNull(getValueViaEnvOrPropertyOrDefault(environment, SONARQUBE_TRANSPORT, "")).toLowerCase(Locale.getDefault());
-    this.isHttpEnabled = transportMode.equals("http") || transportMode.equals("https");
-    this.isHttpsEnabled = transportMode.equals("https");
+    this.isHttpEnabled = isHttpTransport(environment);
+    this.isHttpsEnabled = "https".equals(transportMode(environment));
 
     // Read configuration values.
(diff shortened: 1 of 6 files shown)


nquinquenel:
@damien-urruty-sonarsource Kinda hacky but I did not find something better

src/main/java/org/sonarsource/sonarqube/mcp/configuration/McpServerLaunchConfiguration.java
```diff
@@ -288,6 +287,19 @@ public boolean isHttpEnabled() {
...
+   */
+  public static boolean isHttpTransport(Map<String, String> environment) {
+    var mode = transportMode(environment);
+    return "http".equals(mode) || "https".equals(mode);
+  }
+
+  private static String transportMode(Map<String, String> environment) {
+    return requireNonNull(getValueViaEnvOrPropertyOrDefault(environment, SONARQUBE_TRANSPORT, "")).toLowerCase(Locale.getDefault());
+  }
+
```
damien-urruty-sonarsource, line 302: Probably outside the scope of this PR, but having our own Transport enum would make this cleaner

damien-urruty-sonarsource (approved):
I am not a big fan of the implementation, but it seems we cannot do better with the current SDK. I would have preferred to override the initialize handler and throw a previously recorded error from it, but it's not possible with the current APIs

PR #560, opened 2026-09-11
MCP-621 Download Cloud analyzers from CDN

nquinquenel:
## Summary

- Download SonarQube Cloud analyzers from the CloudFront CDN for all SQC hosts.
- Keep SonarQube Server downloads unchanged and CDN requests anonymous.

----
## Summary by Gitar

- **Plugin download:**
    - Added `SonarCloudCdnPlugins` to download cloud analyzers from the CDN URL pattern
    - Updated `PluginsSynchronizer` to route plugin downloads through the CDN when running on SonarQube Cloud

### DIFF
--- a/src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
@@ -27,6 +27,7 @@ import java.util.stream.Collectors;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.io.FileUtils;
 import org.sonarsource.sonarlint.core.rpc.protocol.common.Language;
+import org.sonarsource.sonarqube.mcp.http.HttpClient;
 import org.sonarsource.sonarqube.mcp.log.McpLogger;
 import org.sonarsource.sonarqube.mcp.serverapi.ServerApi;
 import org.sonarsource.sonarqube.mcp.serverapi.plugins.response.InstalledPluginsResponse;
@@ -90,9 +91,7 @@ public class PluginsSynchronizer {
   }
 
   private void downloadPlugin(String pluginKey, Path localPath, String expectedHash) {
-    try (var response = serverApi.isSonarQubeCloud()
-      ? serverApi.sonarCloudCdnPlugins().downloadPlugin(pluginKey, expectedHash)
-      : serverApi.pluginsApi().downloadPlugin(pluginKey)) {
+    try (var response = getPluginDownloadResponse(pluginKey, expectedHash)) {
       if (response.isSuccessful()) {
         try (var inputStream = response.bodyAsStream()) {
           FileUtils.copyInputStreamToFile(inputStream, localPath.toFile());
@@ -105,6 +104,19 @@ public class PluginsSynchronizer {
     } catch (IOException e) {
       throw new IllegalStateException("Error downloading plugin '" + pluginKey + "'", e);
     }
+  }
+
+  private HttpClient.Response getPluginDownloadResponse(String pluginKey, String expectedHash) {
+    if (serverApi.isSonarQubeCloud()) {
+      var cdnPluginsApi = serverApi.sonarCloudCdnPluginsApi();
+      if (cdnPluginsApi.isAvailable()) {
+        if (expectedHash == null || expectedHash.isBlank()) {
+          throw new IllegalStateException("Cannot download plugin '" + pluginKey + "' from the SonarQube Cloud CDN without an MD5 hash");
+        }
+        return cdnPluginsApi.downloadPlugin(pluginKey, expectedHash);
+      }
+    }
+    return serverApi.pluginsApi().downloadPlugin(pluginKey);
   }
 
   private static void verifyDownloadedPluginHash(String pluginKey, Path localPath, String expectedHash) {
--- a/src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
@@ -90,7 +90,9 @@ public class PluginsSynchronizer {
   }
 
   private void downloadPlugin(String pluginKey, Path localPath, String expectedHash) {
-    try (var response = serverApi.pluginsApi().downloadPlugin(pluginKey)) {
+    try (var response = serverApi.isSonarQubeCloud()
+      ? serverApi.sonarCloudCdnPlugins().downloadPlugin(pluginKey, expectedHash)
+      : serverApi.pluginsApi().downloadPlugin(pluginKey)) {
       if (response.isSuccessful()) {
         try (var inputStream = response.bodyAsStream()) {
           FileUtils.copyInputStreamToFile(inputStream, localPath.toFile());
--- a/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApi.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApi.java
@@ -32,6 +32,7 @@ import org.sonarsource.sonarqube.mcp.serverapi.measures.MeasuresApi;
 import org.sonarsource.sonarqube.mcp.serverapi.metrics.MetricsApi;
 import org.sonarsource.sonarqube.mcp.serverapi.organizations.OrganizationsApi;
 import org.sonarsource.sonarqube.mcp.serverapi.plugins.PluginsApi;
+import org.sonarsource.sonarqube.mcp.serverapi.plugins.SonarCloudCdnPlugins;
 import org.sonarsource.sonarqube.mcp.serverapi.pullrequests.PullRequestsApi;
 import org.sonarsource.sonarqube.mcp.serverapi.qualitygates.QualityGatesApi;
 import org.sonarsource.sonarqube.mcp.serverapi.qualityprofiles.QualityProfilesApi;
@@ -103,6 +104,10 @@ public class ServerApi {
 
   public PluginsApi pluginsApi() {
     return new PluginsApi(helper, isSonarQubeCloud);
+  }
+
+  public SonarCloudCdnPlugins sonarCloudCdnPlugins() {
+    return new SonarCloudCdnPlugins(helper);
   }
 
   public ScaApi scaApi() {
--- a/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApi.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApi.java
@@ -32,7 +32,7 @@ import org.sonarsource.sonarqube.mcp.serverapi.measures.MeasuresApi;
 import org.sonarsource.sonarqube.mcp.serverapi.metrics.MetricsApi;
 import org.sonarsource.sonarqube.mcp.serverapi.organizations.OrganizationsApi;
 import org.sonarsource.sonarqube.mcp.serverapi.plugins.PluginsApi;
-import org.sonarsource.sonarqube.mcp.serverapi.plugins.SonarCloudCdnPlugins;
+import org.sonarsource.sonarqube.mcp.serverapi.plugins.SonarCloudCdnPluginsApi;
 import org.sonarsource.sonarqube.mcp.serverapi.pullrequests.PullRequestsApi;
 import org.sonarsource.sonarqube.mcp.serverapi.qualitygates.QualityGatesApi;
 import org.sonarsource.sonarqube.mcp.serverapi.qualityprofiles.QualityProfilesApi;
@@ -106,8 +106,8 @@ public class ServerApi {
     return new PluginsApi(helper, isSonarQubeCloud);
   }
 
-  public SonarCloudCdnPlugins sonarCloudCdnPlugins() {
-    return new SonarCloudCdnPlugins(helper);
+  public SonarCloudCdnPluginsApi sonarCloudCdnPluginsApi() {
+    return new SonarCloudCdnPluginsApi(helper);
   }
 
   public ScaApi scaApi() {
--- a/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
@@ -54,6 +54,10 @@ public class ServerApiHelper {
     return endpointParams.isSonarQubeCloud();
   }
 
+  public String getBaseUrl() {
+    return endpointParams.baseUrl();
+  }
+
   public HttpClient.Response get(String path) {
     var response = rawGet(path);
     if (!response.isSuccessful()) {
@@ -86,7 +90,11 @@ public class ServerApiHelper {
   }
 
   public HttpClient.Response rawGetAnonymous(String relativePath) {
-    return client.getAsyncAnonymous(buildEndpointUrl(relativePath)).join();
+    return rawGetAnonymousUrl(buildEndpointUrl(relativePath));
+  }
+
+  public HttpClient.Response rawGetAnonymousUrl(String url) {
+    return client.getAsyncAnonymous(url).join();
   }
 
   private HttpClient.Response rawPost(String url, String contentType, String body) {
--- a/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/SonarCloudCdnPlugins.java
+++ b/src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/SonarCloudCdnPlugins.java
@@ -1,0 +1,50 @@
+/*
+ * SonarQube MCP Server
+ * Copyright (C) SonarSource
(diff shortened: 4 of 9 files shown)


src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/SonarCloudCdnPlugins.java
```diff
@@ -0,0 +1,50 @@
...
+    this.helper = helper;
+  }
+
+  public HttpClient.Response downloadPlugin(String pluginKey, String md5) {
+    return helper.rawGetAnonymousUrl(buildDownloadUrl(helper.getBaseUrl(), pluginKey, md5));
+  }
+
+  static String buildDownloadUrl(String baseUrl, String pluginKey, String md5) {
+    var baseUri = URI.create(baseUrl);
+    var host = baseUri.getHost();
```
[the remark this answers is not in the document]
nquinquenel: Addressed in 5982693: the client, facade accessor, and test now use the SonarCloudCdnPluginsApi naming, and PLUGIN_DOWNLOAD_PATH exposes the route template. I kept URI construction instead of UrlBuilder because this URL has encoded path segments and no query parameters; UrlBuilder is designed for query parameters in this codebase.

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
```diff
@@ -90,7 +90,9 @@ private boolean shouldDownload(InstalledPluginsResponse.Plugin plugin) {
   }
 
   private void downloadPlugin(String pluginKey, Path localPath, String expectedHash) {
-    try (var response = serverApi.pluginsApi().downloadPlugin(pluginKey)) {
+    try (var response = serverApi.isSonarQubeCloud()
+      ? serverApi.sonarCloudCdnPlugins().downloadPlugin(pluginKey, expectedHash)
+      : serverApi.pluginsApi().downloadPlugin(pluginKey)) {
```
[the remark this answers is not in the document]
nquinquenel: Fixed in 5982693. On a CDN-backed host, null or blank hashes are rejected before either CDN or Web download is attempted, and existing per-plugin failure isolation handles the error. Tests cover both null and blank values. We intentionally do not fall back for malformed CDN metadata.

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/SonarCloudCdnPluginsApi.java
```diff
@@ -0,0 +1,50 @@
...
+    var baseUri = URI.create(baseUrl);
+    var host = baseUri.getHost();
+    if (host == null) {
+      throw new IllegalArgumentException("SonarQube Cloud base URL must contain a host");
+    }
+    try {
+      return new URI(baseUri.getScheme(), null, "scanner." + host, baseUri.getPort(),
+        "/plugins/" + pluginKey + "/versions/" + md5 + ".jar", null, null).toASCIIString();
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Unable to build SonarQube Cloud CDN plugin URL", e);
```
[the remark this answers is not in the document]
nquinquenel: Fixed in 5982693. CDN use is now limited to the two production Web hosts, sonarcloud.io and sonarqube.us. Other Cloud-mode hosts retain the existing anonymous Web download path, avoiding a regression when scanner.<custom-host> is unavailable. Added routing and availability tests.

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
```diff
@@ -105,6 +106,19 @@ private void downloadPlugin(String pluginKey, Path localPath, String expectedHas
...
+      var cdnPluginsApi = serverApi.sonarCloudCdnPluginsApi();
+      if (cdnPluginsApi.isAvailable()) {
+        if (expectedHash == null || expectedHash.isBlank()) {
+          throw new IllegalStateException("Cannot download plugin '" + pluginKey + "' from the SonarQube Cloud CDN without an MD5 hash");
+        }
+        return cdnPluginsApi.downloadPlugin(pluginKey, expectedHash);
+      }
+    }
+    return serverApi.pluginsApi().downloadPlugin(pluginKey);
+  }
```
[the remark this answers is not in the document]
nquinquenel: Fixed in c2ae2c7. The CDN is now best-effort: missing or blank hashes use the existing Web endpoint, non-success CDN responses are closed before fallback, and runtime transport failures also fall back. Added coverage for missing hashes, HTTP failures, and exceptions.

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/SonarCloudCdnPluginsApi.java
```diff
@@ -0,0 +1,60 @@
...
+    var host = baseUri.getHost();
+    if (host == null) {
+      throw new IllegalArgumentException("SonarQube Cloud base URL must contain a host");
+    }
+    try {
+      return new URI(baseUri.getScheme(), null, "scanner." + host, baseUri.getPort(),
+        PLUGIN_DOWNLOAD_PATH.formatted(pluginKey, md5), null, null).toASCIIString();
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Unable to build SonarQube Cloud CDN plugin URL", e);
+    }
```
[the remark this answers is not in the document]
nquinquenel: Fixed in c2ae2c7. CDN URL construction now accepts only a 32-character hexadecimal MD5, so traversal and other free-form path values are rejected before any CDN request. The installed-plugin hash is modeled as nullable and null or blank values take the Web fallback. Tests cover traversal, malformed hashes, and valid uppercase/lowercase digests.

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
```diff
@@ -105,6 +109,26 @@ private void downloadPlugin(String pluginKey, Path localPath, String expectedHas
...
+    if (serverApi.isSonarQubeCloud()) {
+      var cdnPluginsApi = serverApi.sonarCloudCdnPluginsApi();
+      if (cdnPluginsApi.isAvailable() && expectedHash != null && !expectedHash.isBlank()) {
+        try {
+          var response = cdnPluginsApi.downloadPlugin(pluginKey, expectedHash);
+          if (response.isSuccessful()) {
+            return response;
+          }
+          var statusCode = response.code();
+          response.close();
```
damien-urruty-sonarsource: Do we need to close explictly? I think it will be closed by the caller, as for the happy path

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/response/InstalledPluginsResponse.java
```diff
@@ -17,10 +17,11 @@
 package org.sonarsource.sonarqube.mcp.serverapi.plugins.response;
 
 import java.util.List;
+import jakarta.annotation.Nullable;
 
 public record InstalledPluginsResponse(List<Plugin> plugins) {
 
-  public record Plugin(String key, boolean sonarLintSupported, String filename, String hash) {
+  public record Plugin(String key, boolean sonarLintSupported, String filename, @Nullable String hash) {
```
damien-urruty-sonarsource: Why this change? I don't think hashes are nullable, in SLCORE we always consider them non-null. This adds some checks, so if we can avoid them it's better

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
```diff
@@ -105,6 +109,26 @@ private void downloadPlugin(String pluginKey, Path localPath, String expectedHas
     }
   }
 
+  private HttpClient.Response getPluginDownloadResponse(String pluginKey, @Nullable String expectedHash) {
+    if (serverApi.isSonarQubeCloud()) {
+      var cdnPluginsApi = serverApi.sonarCloudCdnPluginsApi();
+      if (cdnPluginsApi.isAvailable() && expectedHash != null && !expectedHash.isBlank()) {
```
damien-urruty-sonarsource: Isn't the `isSonarQubeCloud` check above enough? It means the way we check if SQC is scattered into different conditions

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/SonarCloudCdnPluginsApi.java
```diff
@@ -0,0 +1,62 @@
...
+    return "sonarcloud.io".equals(host) || "sonarqube.us".equals(host);
+  }
+
+  public HttpClient.Response downloadPlugin(String pluginKey, String md5) {
+    return helper.rawGetAnonymousUrl(buildDownloadUrl(helper.getBaseUrl(), pluginKey, md5));
+  }
+
+  static String buildDownloadUrl(String baseUrl, String pluginKey, String md5) {
+    if (!MD5_PATTERN.matcher(md5).matches()) {
+      throw new IllegalArgumentException("Plugin MD5 must be a 32-character hexadecimal value");
```
damien-urruty-sonarsource: I don't think this should be considered part of the contract. We don't need to check anything IMO, we pass the value we received from the server itself

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApi.java
```diff
@@ -105,6 +106,10 @@ public PluginsApi pluginsApi() {
     return new PluginsApi(helper, isSonarQubeCloud);
   }
 
+  public SonarCloudCdnPluginsApi sonarCloudCdnPluginsApi() {
```
damien-urruty-sonarsource: Nitpick: the ServerApi class mimicks the web API, here it's a different service altogether with a different URL. In SLCORE it was not modeled in the ServerApi on purpose but as a different class

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
```diff
@@ -105,6 +109,26 @@ private void downloadPlugin(String pluginKey, Path localPath, String expectedHas
...
+        try {
+          var response = cdnPluginsApi.downloadPlugin(pluginKey, expectedHash);
+          if (response.isSuccessful()) {
+            return response;
+          }
+          var statusCode = response.code();
+          response.close();
+          LOG.warn("Plugin '" + pluginKey + "' CDN download returned HTTP " + statusCode + "; falling back to the Web API");
```
damien-urruty-sonarsource: I am not sure we should fallback. Julien mentioned they might put a redirect in place, so we might end up doing twice the same request. I bet they will also want to remove the API at some point

damien-urruty-sonarsource (commented):
Looks mostly good, but I think there is an unnecessary check about nullable hashes that we could get rid of

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
nquinquenel: I think if it fails on a CDN response, it's not closing and will leak the resource

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
nquinquenel: I think the redirect was meant as a temporary workaround, but I agree that it may be pointless once they remove the API. I added this as a safe net, but happy to remove it. Also the `analyze_code_snippet` tool is deprecated so this whole logic may even completely disappear at some point...

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/SonarCloudCdnPluginsApi.java
nquinquenel: This was added due to gitar complains, but it's indeed redundant and not a meaningful security check. I'll remove it for simplicity

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/response/InstalledPluginsResponse.java
nquinquenel: This was following the gitar comment above - I think the incentive is that on a SQ regression, they could break the contract and we'd receive null, which would break the CDN request. I can let it fail rather than have a defensive branch, this would simplify the checks and let surface a contract breakage

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
nquinquenel: Given that I'm removing the fallback, this can go. This was following a gitar feedback above, in case some SQC env do not allow for scanner.host . In the ticket, I mentioned that it should be available everywhere (based on SLCORE implem) so it should be fine in any case.

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
```diff
@@ -105,6 +105,13 @@ private void downloadPlugin(String pluginKey, Path localPath, String expectedHas
     }
   }
 
+  private HttpClient.Response getPluginDownloadResponse(String pluginKey, String expectedHash) {
+    if (serverApi.isSonarQubeCloud()) {
+      return sonarCloudCdnPluginsApi.downloadPlugin(pluginKey, expectedHash);
+    }
+    return serverApi.pluginsApi().downloadPlugin(pluginKey);
+  }
```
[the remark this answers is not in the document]
nquinquenel: Expected, scanner.host... should be available on all SQC environments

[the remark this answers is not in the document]
nquinquenel: Correct that Cloud mode is configuration-driven and does not validate the host. We rely on the platform contract: legitimate SQC environments expose scanner.host. Marking an arbitrary endpoint as Cloud is unsupported configuration.

src/main/java/org/sonarsource/sonarqube/mcp/plugins/PluginsSynchronizer.java
```diff
@@ -78,19 +82,15 @@ private boolean shouldDownload(InstalledPluginsResponse.Plugin plugin) {
     if (!Files.exists(localPath)) {
       return true;
     }
-    var expectedHash = plugin.hash();
-    if (expectedHash == null || expectedHash.isBlank()) {
-      return true;
-    }
     try {
-      return !expectedHash.equalsIgnoreCase(computeMd5Hex(localPath));
+      return !plugin.hash().equalsIgnoreCase(computeMd5Hex(localPath));
```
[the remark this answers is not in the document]
nquinquenel: It's not supposed to be null, we let it fail if the contract is broken.

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/response/InstalledPluginsResponse.java
nquinquenel: Removed the null check.

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/plugins/SonarCloudCdnPluginsApi.java
```diff
@@ -0,0 +1,53 @@
...
+  public HttpClient.Response downloadPlugin(String pluginKey, String md5) {
+    return httpClient.getAsyncAnonymous(buildDownloadUrl(baseUrl, pluginKey, md5)).join();
+  }
+
+  static String buildDownloadUrl(String baseUrl, String pluginKey, String md5) {
+    var baseUri = URI.create(baseUrl);
+    var host = baseUri.getHost();
+    if (host == null) {
+      throw new IllegalArgumentException("SonarQube Cloud base URL must contain a host");
+    }
```
damien-urruty-sonarsource, line 44: Maybe not for now, but I think this check should be made way before. It means the user configured something that does not make sense, we should probably throw earlier. The value comes from `mcpConfiguration.getSonarQubeUrl()` so maybe throwing from the `McpServerLaunchConfiguration` class would be good. We could also consider returning a URI from this class instead of a simple String

src/main/java/org/sonarsource/sonarqube/mcp/serverapi/ServerApiHelper.java
```diff
@@ -54,6 +54,10 @@ public boolean isSonarQubeCloud() {
     return endpointParams.isSonarQubeCloud();
   }
 
+  public String getBaseUrl() {
+    return endpointParams.baseUrl();
+  }
```
damien-urruty-sonarsource: This is now dead code

damien-urruty-sonarsource (approved):
Nice!

